PR #146, opened 2020-03-31
Add types package and use PXBlueColor

emclaug2:
- Use the Types package.

Whenever new versions of the Color package are released, the @pxblue/types package might be removed.


huayunh:
Should we list the type package under `doc-it/src/__configuration__/resources`? Or is it too confusing for the PXBlue users?

joebochill:
> Should we list the type package under `doc-it/src/__configuration__/resources`? Or is it too confusing for the PXBlue users?

I would say probably no. Nobody is expected to use the package directly (it's just a dependency of the colors package).

joebochill:
@emclaug2 if you confirm that the beta version of the colors package works as expected i'll publish a latest and we can update that instead of having a direct dependency on the types package.

emclaug2:
@joebochill  The beta version of the @pxblue/colors package currently lists the @pxblue/types package as a transitive dependency.  If someone uses either the @pxblue/colors or @pxblue/colors-branding package, they will have access to the @pxblue/types package as well.  

I tested this in both JS and TS projects with no complications.  

I would not merge this PR yet; doc-it shoud not need need to explicitly list the @pxblue/types package.

PR #155, opened 2020-04-15
Add analytic files and circleCI entry

jeffvg:
- add ga file back to source
- add ga to circleci
- 
gaID


emclaug2:
Will this re-add ga to docit and if so is there a dashboard we can look at to verify it's working?

joebochill:
> Will this re-add ga to docit and if so is there a dashboard we can look at to verify it's working?

Yeah, message me and I'll show you how to get into the GA account.

PR #156, opened 2020-04-17
Update master with analytics enabled

jeffvg:
- 13 total commits including analytics
- 
-


huayunh:
Are master and dev going to share the same Google Analytics token? Will our dev work have a big impact on the GA report?

huayunh:
Looks like the link checker does not like all those android studio links

joebochill:
dev does not have a GA token. GA is only set up for `pxblue.github.io` and `localhost`. These are intentionally separated so that we don't pollute the production analytics with any testing work. And because we do most of our testing in localhost rather than on work already deployed to dev, we set up the analytics for localhost instead of dev.

Link checker is having a problem with a redirect loop on android's site. Since that is the only issue, I'm fine to merge this in overriding the status checks. Hopefully they figure out what they're doing over there, otherwise we can add the link to the ignore pattern for the checker (not ideal).

jeffvg:
should we just try a deeper link in to android studio downloads page>
- https://developer.android.com/studio/#downloads

emclaug2:
> should we just try a deeper link in to android studio downloads page>
> 
> * https://developer.android.com/studio/#downloads

Do you think this will fix the broken link that's failing the pipeline?

jeffvg:
> > should we just try a deeper link in to android studio downloads page>
> > 
> > * https://developer.android.com/studio/#downloads
> 
> Do you think this will fix the broken link that's failing the pipeline? I'm open to either deeper checks or just trying to disable this one rule, whatever is easier.

It might fix it and stop android studio reloading after identifying client OS that populates the download now button

jeffvg:
> > > should we just try a deeper link in to android studio downloads page>
> > > 
> > > * https://developer.android.com/studio/#downloads
> > 
> > 
> > Do you think this will fix the broken link that's failing the pipeline? I'm open to either deeper checks or just trying to disable this one rule, whatever is easier.
> 
> It might fix it and stop android studio reloading after identifying client OS that populates the download now button

verified deep link does not work. Also top level url has same results

PR #157, opened 2020-04-20
Feature/roadmap api

emclaug2:
- Read roadmap data from the API
- Show an empty list (loading) when fetching the roadmap data


src/app/api/index.tsx
```diff
@@ -19,6 +20,10 @@ export const npm = axios.create({
     baseURL: 'https://api.npms.io/v2/',
     timeout: 5000,
 });
+export const roadmap = axios.create({
+    baseURL: 'https://us-central1-doc-it-api.cloudfunctions.net/roadmap/',
```
joebochill, line 24: can we change the baseURL to `'....net/'` and then do `.get('/roadmap')` below?

src/app/pages/Roadmap.tsx
```diff
@@ -85,7 +86,33 @@ export const Roadmap: React.FC = (): JSX.Element => {
...
+    const [roadmap, setRoadmap] = useState<RoadmapBucket[]>([]);
+    const [loading, setLoading] = useState<boolean>(true);
     const theme = useTheme();
+    const loadingGroups = [
+        [1, 2],
+        [1, 2, 3, 4],
+        [1, 2, 3],
+    ];
+
+    useEffect(() => {
```
joebochill, line 98: refer to the useEffect in `BugsButton` for returning a cleanup function that will eliminate functions running on unmounted components.

src/__configuration__/roadmap/index.tsx
```diff
@@ -11,14 +11,14 @@ export type RoadmapItem = {
...
-type RoadmapBucket = {
+export type RoadmapBucket = {
     name: string;
     description: string;
     applies?: FrameworkFilter[];
     items: RoadmapItem[];
 };
 
-export const roadmap: RoadmapBucket[] = [
+export const backupRoadmap: RoadmapBucket[] = [
```
joebochill, line 21: While you are in here, can you update the roadmap data...the workflows in progress are listed as React, but they're actually React Native

src/app/api/index.tsx
emclaug2: Yep, I've updated this

src/__configuration__/roadmap/index.tsx
huayunh: Maybe add list item tag too?

emclaug2: Yep updated these to be reactnative

src/app/pages/Roadmap.tsx
emclaug2: I don't understand exactly what is happening here, not sure how BugsButton will guarantee a mount before the API call comes back.  I copied the logic and moved it over the the Roadmap component.  I figure since the call will always be asynchronous then the component should moment even if the content is cached on the client.

joebochill: It's not about guaranteeing a mount before the call comes back (that's guaranteed based on the fact that we don't make the call until we're mounted)...it's more about checking if the component is still mounted before setting state. If you call setRoadmap (or setLoading) and the component is unmounted (e.g. from navigating to another screen before the asynchronous call resolves) you'll get an error/exception in the console (e.g., "cannot set state on an unmounted component"). The asynchronous functions don't get automatically canceled when the component is unmounted, so we have to check ourselves.

It's probably a little overkill, but it's nice to have the same structure for all of our API calls :)

joebochill:
I updated circleCI so the tests pass now, you just need to merge in the latest from dev to get the temporary fix for the link checker

PR #161, opened 2020-04-23
Roadmap API Refactor

emclaug2:
- Change the API source from Firebase to Github
- Remove the backup roadmap & add an empty state placeholder 

Doc-it using the new API can be viewed here: https://doc-it-api.web.app/

Empty State design as of now: 
image

New API dataflow: https://miro.com/app/board/o9J_ktoYjA4=/


huayunh:
Looks like expo is changing their webpage links

src/__configuration__/roadmap/index.tsx
```diff
@@ -17,201 +17,3 @@ export type RoadmapBucket = {
     applies?: FrameworkFilter[];
     items: RoadmapItem[];
 };
-
```
joebochill, line 20: Since this file is just type definitions now, we should move it out of the configuration folder and into the types area.

src/app/api/index.tsx
```diff
@@ -21,7 +21,7 @@ export const npm = axios.create({
     timeout: 5000,
 });
 export const roadmap = axios.create({
-    baseURL: 'https://us-central1-doc-it-api.cloudfunctions.net',
+    baseURL: 'https://raw.githubusercontent.com/pxblue/pxb-database/master/database/doc-it',
```
joebochill, line 24: is there a need to have the extra /database directory or could we have /doc-it at the root level in the repository?

src/app/api/index.tsx
```diff
@@ -90,8 +90,8 @@ export const getNpmVersion = async (packageName: string): Promise<string | undef
 
 export const getRoadmap = async (): Promise<RoadmapBucket[] | undefined> => {
     try {
-        const response = await roadmap.get('/roadmap');
-        if (response && response.status === 200) return response.data;
+        const response = await roadmap.get('/roadmap.json');
```
joebochill, line 93: suggesting we version these by calling them something like `R16Roadmap.json`

src/app/pages/Roadmap.tsx
```diff
@@ -210,6 +211,14 @@ export const Roadmap: React.FC = (): JSX.Element => {
                     </div>
                 )}
 
+                {!loading && filteredBuckets.length === 0 &&
+                    <EmptyState
```
joebochill, line 215: Per our design pattern, this empty state component should be centered (horizontally and vertically) in its parent (the page)

src/__configuration__/roadmap/index.tsx
emclaug2: Yep this makes sense, I've moved the Roadmap types into the type3s area.

src/app/api/index.tsx
emclaug2: I can update the database and start versioning out our roadmaps.

src/app/api/index.tsx
emclaug2: I was thinking this pxb-database repo could be extended beyond doc-it and potentially used for other things in the future.  That's why I put a doc-it bucket for now.  I think it makes sense to leave the doc-it folder for now.

joebochill: Yes, totally agree about keeping the doc-it folder. Just wondering if we should go from:
`pxb-database/database/doc-it` to
`pxb-database/doc-it`

joebochill: or maybe we call the folder 'deployed' or something. I just looked at the database repo, so I understand why there's an extra folder now.

src/app/pages/Roadmap.tsx
emclaug2: I centered the emptyState, but I did it in a hacky way because I couldn't get it to work using height 100% and flex.

joebochill: It's a little bit of a pain because of the way we have that footer configured. If you want to get it close and relatively clean:

- Only render the `<PageContent>` component around the actual contents (don't wrap the empty state component in it)
- on the empty state wrapper:
    ```
    min-height: calc(100vh - 112px);
    display: flex;
    align-items: center;
    justify-content: center;
```

joebochill: actually, your hacky way looks nicer because the footer doesn't disappear. But I would do 28vh (you're missing the h) always and not do the 15vh for small screen

emclaug2: The 28vh on mobile viewports looked pretty off, I'll show a picture of what I was seeing:

emclaug2: This is what an iphone 6/7/8 would look like with a vh of 20: 
image

emclaug2: I think 22vh on mobile is the best. 
image

src/app/api/index.tsx
emclaug2: I updated the output folder to `deployed` instead of `database` and update doc-it to pull it down.  Surprisingly github api has been very responsive.  I think now that there aren't any service issues this might be what to expect.   At least for URL changes the updates have been instant.

src/app/pages/Roadmap.tsx
joebochill: It all kind of depends on how big the screen is, but this should be fine for now

PR #165, opened 2020-05-06
Swap out copyrighted material

huayunh:
- Remove and replace all the image assets in the placeholder folder
- Add a design system image (so that I can finally delete it from my desktop)
- Minor headline adjust


huayunh:
Have to say that it bothers me when the "for developers" button is not using javascript code as its background

huayunh:
@joebochill Same for the designer pic :P I drew on a scratch paper for docit wireframing

PR #177, opened 2020-05-21
Add gcse verification meta tag

huayunh:
- Add a meta tag to index.html to make our site appear in a regular google search
- Shall be pushed to master branch to have this go through

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,21 +1,19 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8" />
-    <link rel="icon" href="%PUBLIC_URL%/pxblue.png" />
-    <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <meta name="theme-color" content="#007bc1" />
-    <meta
-      name="description"
-      content="Documentation site for the PX Blue design system."
-    />
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/images/pxblue192.png" />
-    <!--
+    <head>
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -1,21 +1,19 @@
...
-    />
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/images/pxblue192.png" />
-    <!--
+    <head>
+        <meta charset="utf-8" />
+        <link rel="icon" href="%PUBLIC_URL%/pxblue.png" />
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
+        <meta name="theme-color" content="#007bc1" />
+        <meta name="description" content="Documentation site for the PX Blue design system." />
+        <meta name="google-site-verification" content="iuEyJbuvDYHf_Ksbyb7W4CzOpQVTzoKWEPzZN4oOluk" />
```
joebochill, line 9: Do we need one for master and one for dev, or does it not matter that they are different sites with the same ID?

huayunh: I am trying to verify develop.pxblue as the owner for pxblue.github.io only. While we can verify px-blue.github.io, I don’t think it is a good idea to have them index our dev site.

joebochill: Sounds good. Just wondering if it will cause a problem having the id exist in multiple places, or if google is just going to look in that one location for the code and ignore everywhere else.

huayunh: Maybe we can add to ci to modify the robots.txt for each site differently

PR #179, opened 2020-05-22
Color, RNP instructions, Google site verification meta tag

huayunh:
- Colors page 
- RNP instructions
- Google site verification


joebochill:
I don't want to do a master deployment with the new RN instructions until we've published the new versions of the themes and component library (since the instructions won't work).

Can you do this Google Meta Tag change as a hotfix into master instead of dev - > master?

huayunh:
No problem

PR #199, opened 2020-06-11
Designer Environment Set Up

huayunh:
- Incorporate designer content to docit.


src/docs/design/intro.mdx
```diff
@@ -4,6 +4,10 @@ import { Divider } from '../../app/components';
 
 Using the PX Blue design system as a designer can greatly improve your efficiency and the quality of your products. PX Blue has many patterns and guidelines (color usage, typography, iconography, etc.) to address the most common elements and interactions that you might encounter when designing a product. By following these recommendations and using pre-defined resources from PX Blue, your time is free to focus on the elements that make your project unique. In general, we recommend trying to use PX Blue for 80% of your product's needs.
 
+## Environment
+
+We currently actively host and maintain our design assets via Figma. Figma is a powerful prototyping tool that facilitates team collaborations and design assets sharing. It works on both browsers and, if you prefer, a desktop app (Mac and PC). You may [sign up here](https://www.figma.com/).
```
daileytj, line 9: can we remove 'currently' or 'actively'... seems redundant to have both.

joebochill, line 9:
```suggestion
We host and maintain our design assets through Figma. Figma is a powerful prototyping tool that facilitates collaboration and sharing of design assets between teams. It can be used on the web, and they also provide a desktop app for Mac and Windows. In order to use Figma, you will need to [sign up](https://www.figma.com/) for an account.
```

src/docs/design/intro.mdx
```diff
@@ -13,9 +17,9 @@ PX Blue has design/style guidelines for:
 -   [Color Usage](/style/color)
 -   [Themes](/style/themes)
 
-In addition to these guidelines, we also provide a [Figma](https://www.figma.com/) library that gives you access to pre-built styles and assets that you can drag and drop to build designs for your applications. [Contact us](/community/contactus) to request access to this library. A future release of PX Blue will include an easy-to-use plugin that you can include in your Figma projects to keep you up-to-date with the latest changes and improvements.
+In addition to these guidelines, we also provide [an easy-to-use plugin](https://www.figma.com/community/file/852558784352181868) that you can include in your Figma projects to keep you up-to-date with the latest changes and improvements. You are encouraged to duplicate this file to your own draft space and build your project on top of it.
```
joebochill, line 20: Is this a plugin? I reworded this in my PR because I wasn't sure if it works like this. If somebody forks the sticker sheet and we make an update to it, do they get the changes?

huayunh: It is called a "plugin" by Figma, but I think it is probably closer to a "forkable repository". 

### Here's my own understanding: 
I do not think they will automatically get the latest version when the designer "push" it. Every time a designer "fork" (click that duplicate button), they will get a copy of the most recent version of it in their own draft space. They can play with it on their own. If we make any updates to the community, they do not automatically receive the latest update. 

Next time they build another project, they will fork again, which will generate a new draft with the latest version, but nothing will be updated to the previous projects.

joebochill: If Figma calls it a plugin we can call it a plugin. But I wouldn't say 'to keep you up to date with the latest changes and improvements' since it won't do that. It might be better to say 'to jump start your project with pre-defined UI elements' or something like that.

PR #205, opened 2020-06-25
Update Form Validation Doc (2)

huayunh:
- Incorporate a little feedback from Ian
- Add more graphics that can "go with" the content


huayunh:
I'd like to show this to Ian after some rewording help :P

src/docs/patterns/forms.mdx
```diff
@@ -82,3 +95,78 @@ In some specific situations, you may need to defer verification or perform it fo
...
+    gridImageProps={{ sm: 12 }}
+    caption={
+        'If you have only one input field, and this field is verified on change, disable the submit button until the user entered the desired input.'
+    }
+/>
+<ImageGrid
+    images={[SubmitButtonAutoSubmitImage]}
+    gridImageProps={{ sm: 12 }}
+    caption={
+        'For the verification code screen, you can auto-submit the code once the user input reaches the desired length (2). After the verification code is recognized (3), the screen pauses for a short moment before auto-transit to the next page. The verify button is preserved for consistency.'
```
joebochill, line 136:
```suggestion
        'For the verification code screen, you can auto-submit the information once the user input reaches the required length (#2). After the verification code is recognized (#3), the screen pauses for a short moment before automatically moving to the next screen. The verify button is preserved for consistency.'
```
"The verify button is preserved for consistency" - What does this mean? From the description it sounds like the verification button is not used at all, so why even have one?

This workflow seems like a special case. Does it warrant being called out at all? And if so, should it be more prominent than just an image caption?

src/docs/patterns/forms.mdx
```diff
@@ -82,3 +95,78 @@ In some specific situations, you may need to defer verification or perform it fo
...
+    }
+/>
+
+If at least one text field uses the 'verify on blur' strategy, do not disable the submit button. Otherwise, the mobile user will have to disable their keyboard first to have their input field verified, before they can click the submit button. The only exception is if there are required input fields.
```
joebochill, line 140: This is only relevant on mobile right? Desktop users shouldn't be disabling their keyboards 😛 

If above we are saying that onBlur should be the most common strategy, this scenario will be a very common one.  What is the proper way of doing this?

Should we mention somewhere about being able to submit the form with the Enter key (for desktop users)?

src/docs/patterns/forms.mdx
```diff
@@ -82,3 +95,78 @@ In some specific situations, you may need to defer verification or perform it fo
...
+
+<ImageGrid
+    images={[SubmitButtonOnBlurImage]}
+    gridImageProps={{ sm: 12 }}
+    caption={'The submit button is disabled until the user has entered something for all the required fields.'}
+/>
+
+When you have a long form (more than three text fields) with a lot of text fields and the user forgets a required field, do not auto disable the submit button -- it will take them some effort to check what they are missing. Instead, allow the user to click the submit button, and highlight the errors for them.
```
joebochill, line 148: Seems like a lot of conflicting advice. Above we say "use onBlur for most things", then we say "if you're using on blur, don't disable the submit button", and then we say "actually, disable the submit button if you have required fields", and now we say "but if you have a long form, don't disable it even if you have required fields".

Maybe we could have some small example sections instead of trying to be too generic. E.g., "This form has two fields, one of which is required. THIS IS THE BEHAVIOR YOU SHOULD IMPLEMENT IN THIS SITUATION." "This is a multi-page form (first page) and all fields are optional. THIS IS THE BEHAVIOR YOU SHOULD IMPLEMENT IN THIS SITUATION."

joebochill, line 148:
```suggestion
When you have a long form (more than three text fields) and the user forgets a required field, do not disable the submit button and force the user to search for what they are missing. Instead, allow the user to click the submit button, and highlight the errors for them.
```

src/docs/patterns/forms.mdx
```diff
@@ -82,3 +95,78 @@ In some specific situations, you may need to defer verification or perform it fo
...
+We recommend using the following components to build this pattern:
+
+**Angular** - Angular Material
+
+-   [Form Field](https://material.angular.io/components/form-field/overview)
+-   [Input](https://material.angular.io/components/input/overview)
+
+**React** - Material-UI
+
+-   [Text Field](https://material-ui.com/components/text-fields/#text-field)
```
joebochill, line 172: I think this goes without saying...there's not really an alternative

src/docs/patterns/forms.mdx
```diff
@@ -82,3 +95,78 @@ In some specific situations, you may need to defer verification or perform it fo
...
+## Submit the Form
+
+Usually, a form will have a submit/next/finished button at the end. This button is typically placed at the bottom right corner of the form. If you are designing for a right-to-left language, this button should be at the bottom left.
+
+<ImageGrid
+    images={[SumbitButtonLocationImage]}
+    gridImageProps={{ sm: 12 }}
+    caption={[
+        '1: On desktops, the button is placed in the bottom corner of the form.',
+        '2: On mobile devices, the button is occluded by the native keyboard.',
```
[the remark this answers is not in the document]
huayunh: I think in this case both are valid, although I was trying to put a stronger emphasis on the second point.

hmm... On mobile devices, if the user cannot submit bc the next button is disabled, then the native keyboard will be dismissed when they click the next button, I think.

src/docs/patterns/forms.mdx
huayunh: For the consistency -- if a submit button was there for all the other patterns except for this one, it might feel strange. Although in WAS, instead of "submit", it was a wide "resend confirmation email" button instead. Maybe I should change it -- want to get some big reactions from Ian.

I in fact wanted to put it to login workflow, but we have more contents on the form validation side at the moment, it seems better to have it here

src/docs/patterns/forms.mdx
huayunh: Good idea for the enter key :) 

On blur for most scenarios; thus do not disable the submit button in most cases. 

Eh, I should just have a bullet list of do-disable-submit-button scenarios (instead of the other way)

src/docs/patterns/forms.mdx
huayunh: yeah, it is complicated. I'd need a huge flow chart for that 😛

src/docs/patterns/forms.mdx
huayunh: Yep, just a convenient link for them.

huayunh:
I did this mostly as an experiment for design pattern refactoring (which I am looking forward to). Yet this is going really wild. I think this PR will stay draft for a long time.

src/docs/patterns/forms.mdx
joebochill: I would just leave it out. We've got the recommended components section on the getting started guide. If we're linking out to specific components from all of out patterns it could get crazy. Or maybe we do it as a "More Resources" kind of section with one of our action cards to link out to assorted other things.

joebochill:
> I did this mostly as an experiment for design pattern refactoring (which I am looking forward to). Yet this is going really wild. I think this PR will stay draft for a long time.

I think the content is great. Just needs to be solidified a little bit and answer some open questions. I'm excited for it to eventually get to doc-it.

huayunh:
image
I added a design specs file. Note that the illustrations in the behavior section are not following the dimensions specified here -- we need to revisit these when we are actually refactoring the design patterns.

huayunh:
jasmine nuked their own github page?

src/docs/patterns/forms.mdx
```diff
@@ -74,14 +103,97 @@ Some fields should verify the user input as it changes. The most common case for
...
+<ImageGrid
+    images={[SumbitButtonLocationImage]}
+    regularWidth
+    caption={[
+        '1: On desktops, the button is placed in the bottom corner of the form.',
+        '2: On mobile devices, the button is replaced by a button on native keyboard.',
+    ]}
+/>
+
+On mobile, if there are multiple input fields on the screen, the keyboard button should say "next", unless you are on the last item of the page, in which case the keyboard button should say "done" or "submit". In this situation, depends on if you want your user to double check their input, either dismiss the keyboard upon clicking the button (blur away from the input field), or transition the UI to the next page (submit the form directly).
```
joebochill, line 131:
```suggestion
On mobile, if there are multiple input fields on the screen, the keyboard button should say "next", unless you are on the last item of the page, in which case the keyboard button should say "done" or "submit". This button can either submit the form or simply dismiss the keyboard (if you want users to have a chance to review their information first)..
```

src/docs/patterns/forms.mdx
```diff
@@ -74,14 +103,97 @@ Some fields should verify the user input as it changes. The most common case for
...
+    images={[NextButtonUsageImage]}
+    regularWidth
+    caption={[
+        ,
+        '1: The user entered a valid input. Clicking "next" will autofocus the user to the next text field.',
+        '2: The user entered an invalid input. Clicking "next" will display the error message but will not autofocus to the next text field.',
+    ]}
+/>
+
+For screens like email verifications where there is only one input field with a specific length, and the user will almost never need to go back and double check their input, the submit button can be omitted.
```
joebochill, line 143:
```suggestion
For screens such as email verification where there is a single field with a known required length, the submit button can be omitted.
```

src/docs/patterns/forms.mdx
```diff
@@ -74,14 +103,97 @@ Some fields should verify the user input as it changes. The most common case for
...
+    images={[SubmitButtonAutoSubmitImage]}
+    regularWidth
+    caption={
+        'For the verification code screen, you can auto-submit the information once the user input reaches the required length (#2). After the verification code is recognized (#3), the screen pauses for a short moment before automatically moving to the next screen.'
+    }
+/>
+
+### Disabling the Submit Button
+
+In most cases, the submit button **should not** be disabled. The users should not be forced to search for what they are missing, or what they did wrong.
```
joebochill, line 155: Don't they still have to search if the submit button is enabled? Won't the error messages be there even when the submit button is disabled?

src/docs/patterns/forms.mdx
```diff
@@ -74,14 +103,97 @@ Some fields should verify the user input as it changes. The most common case for
...
+
+<ImageGrid
+    images={[SubmitButtonOnClickImage]}
+    regularWidth
+    caption={
+        'Do not disable the submit button. Instead, use error text to highlight the invalid input after the user clicks the button.'
```
joebochill, line 161: error text next to the button is a little weir...i haven't seen that anywhere

src/docs/patterns/forms.mdx
```diff
@@ -74,14 +103,97 @@ Some fields should verify the user input as it changes. The most common case for
...
+    }
+/>
+
+This rule has an exception: on desktop applications, when there are less than four text fields, and all the fields are verified on change, we recommend you disable the submit button to indicate that there are some errors in the user input (invalid value, missing a required field).
```
joebochill, line 165:
```suggestion
> **Exception**: on desktop applications, when there are fewer than four text fields, and all the fields are verified on change, we recommend you disable the submit button to indicate that there are some errors in the user input (invalid value, missing a required field).
```

joebochill: why do we make the distinction here (why is this the exception)?

huayunh: Screens are bigger, can see more things at once

src/docs/patterns/forms.mdx
huayunh: Applies to people who skipped a field.

src/docs/patterns/forms.mdx
huayunh: Let me ask Ian

huayunh:
I am going to comment out the Developer section for now. However, in the future, I think it is beneficial to include that. If we have a developer section everywhere else, I'd have the developers section there just for consistency.

src/docs/patterns/forms.mdx
joebochill: Can we say that in the description?

huayunh:
I don't feel like working on this any further. I've incorporated most of Ian's feedback into this -- so there should be no more "bugs" here, only a couple of "features" that Ian thinks would be nice to have. I will update Jira to include Ian's comments when we formally look into these patterns.

src/docs/patterns/forms.mdx
```diff
@@ -80,14 +114,119 @@ Some fields should verify the user input as it changes. The most common case for
...
+        images={[SubmitButtonOnBlurImage]}
+        regularWidth
+        caption={'The submit button is disabled until the user has entered something for all the required fields.'}
+    />
+</React.Fragment>
+<Divider />
+
+# Specs
+
+<ImageGrid images={[Specs1Image, Specs2Image, Specs3Image]} regularWidth />
```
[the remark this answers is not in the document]
joebochill: Sometimes the space between fields can be a little weird. I can't remember which is which, but some libraries have the helper text space present even when no helper text is defined. So if we're specifying 24 between fields, that might mean a top margin of 0 if the upper field has no helper text. I don't know if anything needs to change here, just wanted to point that out.

src/docs/patterns/forms.mdx
```diff
@@ -1,36 +1,70 @@
...
 When using a mixture of optional and required fields in a form, the inputs should be labeled appropriately.
 
 If most of the fields are required, you should call out the fields that are optional. These fields should be labeled with an appropriate descriptor, followed by "optional" in parentheses.
 
-Conversely, if most of your fields are optional, you should only call out the required fields. These should be identified by an asterisk (\*) after the field label. There should be a descriptor nearby that indicates that a \* means the field is required.
+Conversely, if most of your fields are optional, you should only call out the required fields. These should be identified by an asterisk (\*) after the field label. Optionally, you can add a descriptor nearby that indicates that a \* means the field is required.
+
+## Form Length
+
+The form should not be too lengthy -- no more than the screen height on the mobile device, and no more than half of the desktop screen. Consider breaking long forms into subsections or multiple pages using tabs and [steppers](/patterns/steppers).
```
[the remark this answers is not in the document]
huayunh: Subsections are mostly so that people will have a chance to breathe in the middle of a long form, and make it easier to skim read/navigate.

src/docs/patterns/forms.mdx
```diff
@@ -48,17 +82,17 @@ Error messages should be succinct and should not provide information that could
```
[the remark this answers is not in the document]
huayunh: I suggest taking that out. Asked Ian and he said it is weird and not used anywhere in the design practice.

src/docs/patterns/forms.mdx
```diff
@@ -80,14 +114,119 @@ Some fields should verify the user input as it changes. The most common case for
...
+## Submitting the Form
+
+Typically, forms should have a button to submit the data or continue to the next part of the form (for long forms). Depending on what happens next, the button should be labeled "next", "finish", or similar. The button should be located at the bottom right corner of the form (bottom left for right-to-left languages).
+
+<ImageGrid
+    images={[SumbitButtonLocationImage]}
+    regularWidth
+    caption={[
+        '1: On desktops, the button is placed in the bottom corner of the form.',
+        '2: On mobile devices, the button is replaced by a button on native keyboard.',
```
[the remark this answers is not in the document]
huayunh: I guess there was some misunderstanding on my end when you mentioned "replace" earlier. What I was trying to say was, on mobile, there are both a submit button at the end of the form ("UI Submit Button"), and a next/submit button on the native keyboard ("Keyboard Submit Button"), depends on which field you are focusing on. "Part of the UI submit button's functionality is replaced by the keyboard submit button" is what I am trying to say here.

src/docs/patterns/forms.mdx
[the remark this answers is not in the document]
huayunh: I think the only thing that a desktop might differ from the mobile example is the form margin with regarding to its container elements -- I will need to double-check on this.

src/docs/patterns/forms.mdx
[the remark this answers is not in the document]
huayunh: Umm... see if this is more reasonable for you:

```suggestion
The form should not be too lengthy -- no more than the smallest screen height for your target devices. Consider breaking long forms into subsections or multiple pages using tabs and [steppers](/patterns/steppers).
```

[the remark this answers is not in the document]
joebochill: Instead of making it two rules, would it be easier to just say something like:
"Keep your forms short where possible. If your forms are larger than the height of your target device, consider breaking them into sections or multiple pages to improve the user experience"

src/docs/patterns/forms.mdx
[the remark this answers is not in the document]
joebochill: On mobile, the software keyboard also contains a submit button for quick access

joebochill:
@huayunh This PR has been open for a while and gotten pretty large. Are we comfortable saying that it's good enough for now and merging it in? We can always follow up with additional tweaks...if it's deployed to dev it will be easier for Ian to provide comments on it too.

huayunh:
@joebochill I'd love to make those final tweaks you mentioned in the PR above, then call it done
Can throw the remaining issues in the backlog overhaul story

src/docs/patterns/forms.mdx
```diff
@@ -80,14 +114,119 @@ Some fields should verify the user input as it changes. The most common case for
...
+    regularWidth
+    caption={
+        'For the verification code screen, you can auto-submit the information once the user input reaches the required length (#2). After the verification code is recognized (#3), the screen pauses for a short moment before automatically moving to the next screen.'
+    }
+/>
+
+When the user attempts to submit by clicking on the submit button on their screen (desktop) or on their native keyboard (mobile), all the invalid fields should be highlighted with red error text.
+
+<ImageGrid
+    images={[dismissKeyboardErrorAutoscrollImage, dismissKeyboardErrorTextImage]}
```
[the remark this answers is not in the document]
huayunh: Not necessarily in the auto-scroll example.

src/docs/patterns/forms.mdx
```diff
@@ -1,36 +1,70 @@
...
+        react={
+            'https://codesandbox.io/s/github/pxblue/react-design-patterns/tree/master?file=/src/pages/form-validation/index.tsx&initialpath=form-validation'
+        }
+    />
+    <Divider />
+</React.Fragment>
+
+# Behaviors
 
 ## Optional vs. Required Fields
```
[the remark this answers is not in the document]
huayunh: Deleted. I was sooooo confused with what it means by "descriptor"; after double-check with Material's documentations, I realized that you mean the assistive text / helper text.

huayunh:
I thought this would be the easiest one, yet still, I managed to spend nearly 1 week on this in total

joebochill:
The two highcharts links are broken...looks like they did an update on their site. Their own internal link is also broken though.

If we want, for now, we can update those to just be https://shop.highsoft.com/ or https://shop.highsoft.com/developer-license

src/docs/patterns/forms.mdx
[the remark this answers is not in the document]
joebochill: Yeah, I just meant the entire part about how to label optional/required fields. Pretty sure they explain this in detail and we weren't saying anything different.

PR #213, opened 2020-06-29
Feature/add section link

### LINKED ISSUE #208: Add a section link to headlines

try adding something like this to all the headlines. On click, the link should be copied and pasted into your clipboard

image

image

#### Describe the current behavior

It does not.

#### Is this request related to a current issue?

No

huayunh:
Fixes #208

On hover: 
image

On click: 
image


jeffvg:
Should we always show copy paper clip to copy rather then on hover? Mostly for mobile

huayunh:
Open to suggestions. Material Design has its paper clip always shown.

src/__configuration__/markdown/copy.ts
```diff
@@ -0,0 +1,24 @@
+// https://stackoverflow.com/questions/400212/how-do-i-copy-to-the-clipboard-in-javascript
```
joebochill, line 1: does this file belong in the config folder, or is it more of a utility or lib?

src/__configuration__/markdown/markdown.css
```diff
@@ -1,6 +1,22 @@
 .markdownH1:first-child {
     margin-top: 32px;
 }
+.markdownH1:first-child .MuiSvgIcon-root {
+    display: none;
```
joebochill, line 5: why no icons for first one? Click still works.

huayunh: It's usually the H1 on the very top, which feels meaningless to add a link icon. Looks like I better figure something out other than the CSS hack? Or I should just allow the first H1 to also be clicked for consistency

joebochill: I think it's fine for now...more just curious if there was a rationale. Keeping the page title clean seems like a good enough reason for me

src/__configuration__/markdown/copy.ts
huayunh: Moved to shared/utilities.tsx

PR #221, opened 2020-07-07
Add additional roadmap filters

joebochill:
For PXBLUE-1370.

- Add filters to roadmap toolbar for Type and Status
- Refactor some of the filtering and empty state logic
- Rename 'applies' to 'framework'
- Add CLI and Workflow to the resources page

#### Screenshots:
image


joebochill:
Codesandbox is having some issues today, hence the broken link check

huayunh:
follow up bug @joebochill 
1. set the framework filter to anything but the "all framework" filter
2. set type to design
3. empty state shows up

should clear the framework filter when the type is set to design, otherwise looking good 

would be even better if you enable scrubbing left and right for the filters on mobile

joebochill:
@huayunh see #222

PR #222, opened 2020-07-10
Fix Roadmap Filtering Behaviors

joebochill:
For @huayunh 

- Don't filter by framework when design is selected
- Let the filter toolbar scroll horizontally on smaller screens


huayunh:
image
Would be even more awesome if there is a 16px padding to the right of "any status". Minor stuff. Looking good otherwise.

joebochill:
> Would be even more awesome if there is a 16px padding to the right of "any status". Minor stuff. Looking good otherwise.

This is, turns out, an annoyance with how CSS works (https://stackoverflow.com/questions/26888428/display-flex-loses-right-padding-when-overflowing). I put in a `<Spacer/>` hack to fix it (not ideal, but it was quick and dirty).

PR #226, opened 2020-07-23
Add app bar design spec picture

huayunh:
- Add design specifications & developer section
- Linked to relevant MD design spec, MUI & Angular Material API documents
- Checked with Ian

#### Screenshots:

image


src/docs/patterns/appbar.mdx
```diff
@@ -35,3 +37,37 @@ In some situations, you may want to present more information in an app bar, such
...
+
+For additional instructions, refer to [Material Design's design specs on "App bars: top"](https://material.io/components/app-bars-top#specs).
+
+## Mobile
+
+<ImageGrid images={[Specs1Image]} regularWidth />
+
+## Desktop and Tablet
+
+If certain dimensions are not specified, refer to the dimensions suggested in the mobile section above.
```
joebochill, line 53: There are no dimensions suggested in the mobile section above

src/docs/patterns/appbar.mdx
```diff
@@ -35,3 +37,37 @@ In some situations, you may want to present more information in an app bar, such
...
+
+<ImageGrid images={[Specs2Image]} regularWidth />
```
joebochill, line 55: For the appbar left icon, google seems to recommend 16px left and 32px right padding (which is what we've been implementing in our components):
image

However, their examples seem to be a mix of 16/32 and 24/24. But I didn't see any 24/32.

My preference would be to stick with 16/32 across the board to maintain the 72px keyline throughout the app. Angular Material and Material UI also both use 16px on the left side out-of-the-box, so if we're recommending 24, then every time someone uses an Appbar they're going to have to override the default styles.

This also begs the question...if we're following Material specs, do we even need to have a spec image here, or do we just refer them to the Material documentation?

src/docs/patterns/appbar.mdx
```diff
@@ -35,3 +37,37 @@ In some situations, you may want to present more information in an app bar, such
...
+
+<Divider />
+
+# Design Specifications
+
+For additional instructions, refer to [Material Design's design specs on "App bars: top"](https://material.io/components/app-bars-top#specs).
+
+## Mobile
+
+<ImageGrid images={[Specs1Image]} regularWidth />
```
joebochill, line 49: Do we need to mention color here? Inherit is the default CSS behavior and does it make sense to a designer?

Should we instead say something like "white[50] on dark backgrounds and black[500] on light backgrounds"?

src/docs/patterns/appbar.mdx
```diff
@@ -35,3 +37,37 @@ In some situations, you may want to present more information in an app bar, such
...
+-   [@angular/material/toolbar](https://material.angular.io/components/toolbar)
+
+**React**:
+
+-   [@material-ui/core/AppBar](https://material-ui.com/api/app-bar/)
+-   [@material-ui/core/Toolbar](https://material-ui.com/api/toolbar/)
+
+**React Native**:
+
+-   [@pxblue/react-native-components/Header](https://pxblue-components.github.io/react-native/?path=/info/components-documentation--header)
```
joebochill, line 73: I would like for us to come up with a more interesting way to present this information instead of a bulleted list...maybe another cool callout card of some kind - we can build a shareable component.

src/docs/patterns/appbar.mdx
```diff
@@ -35,3 +37,37 @@ In some situations, you may want to present more information in an app bar, such
...
+## Desktop and Tablet
+
+If certain dimensions are not specified, refer to the dimensions suggested in the mobile section above.
+
+<ImageGrid images={[Specs2Image]} regularWidth />
+<Divider />
+
+# Developers
+
+We recommend using the following components to build this pattern:
```
joebochill, line 60: I would shy away from saying 'recommend' and instead say "Use the following components to implement this pattern:"

src/docs/patterns/appbar.mdx
joebochill: Where's Ionic?

src/docs/patterns/appbar.mdx
```diff
@@ -35,3 +37,37 @@ In some situations, you may want to present more information in an app bar, such
 <DemoCard repository={'search-bar'} angular react={'https://codesandbox.io/s/github/pxblue/react-design-patterns/tree/master?file=/src/pages/app-bar/search-bar/index.tsx&initialpath=search'} ionic reactNative />
 
 A very common pattern seen in many types of applications is the ability to search for data on a page. This is achieved by placing a search icon/button in the app bar. When this button is clicked, a white Search Bar slides in from the right to cover the default toolbar. As the user enters text into the search field, the data is filtered in real-time.
+
+<Divider />
+
+# Design Specifications
+
+For additional instructions, refer to [Material Design's design specs on "App bars: top"](https://material.io/components/app-bars-top#specs).
```
joebochill, line 45: This is the same link as in the callout card above...should we move that down here? Do we need it in both places?

src/docs/patterns/forms.mdx
```diff
@@ -205,6 +205,8 @@ either remvoe it, or find a better way to picture this problem (like those do's
 
 # Design Specifications
 
+For additional instructions, refer to [Material Design's design specs on "Text fields"](https://material.io/components/text-fields#specs).
```
joebochill, line 208: Same question as above...do we need this here if we already link out to it in the material card above? Should we move the material card down here instead?

src/docs/patterns/appbar.mdx
huayunh: It works like this: 
- look at desktop & tablet dimension
- has no dimension 
- look at mobile
- has no mobile dimension
- look at the Material Design spec
- Aha, here it is

So kinda like some inherit thing... or should we just "import" the material dimensions, so that people don't have to look here and there?

src/docs/patterns/appbar.mdx
huayunh: I must have it in my mind that docit is using 24/32 and thus I recommended 24 as left padding...

I can remove that 24/32, and claim docit is getting some "special treatment" :P

src/docs/patterns/appbar.mdx
huayunh: I should probably just remove it unless there are certain colors not immediately obvious.

src/docs/patterns/appbar.mdx
huayunh: Change it to Angular & Ionic?

src/docs/patterns/appbar.mdx
huayunh: Adding to that, I think the design spec is not only for designers... I am trying to find the best language for both designers and developers.

src/docs/patterns/appbar.mdx
huayunh: Yes. I think we should not make our users scroll all the way to the top, but just offer them handy... Or, we can simply "import" the material design dimensions, but replacing the components with pxb app bars.

src/docs/patterns/appbar.mdx
huayunh: > to come up with a more interesting way to present this information instead of a bulleted list
I can add this into the backlog as part of the design overhauls... our docs are already full of bullet lists, so perhaps it is fine here?

src/docs/patterns/appbar.mdx
joebochill: Doc-It got special treatment - I can't remember what the exact issue was, but it was something to do with the negative margins of the GridList on a smaller width screen...I think the options were to refactor the grid list spacing, or just bump up the toolbar padding to 24 and we went with the easier approach.

src/docs/patterns/appbar.mdx
joebochill: I would probably leave this out like you said unless it's a non-obvious color.

src/docs/patterns/appbar.mdx
joebochill: > Change it to Angular & Ionic?

It's been a while, I can't remember if we use the mat-toolbar for Ionic apps or if we use the Ionic toolbar...it might be a mix.

joebochill: > I can add this into the backlog as part of the design overhauls... our docs are already full of bullet lists, so perhaps it is fine here?

That's fine, I just think bulleted lists are boring for this kind of information (and they take up a bunch of space when we're showing stuff for all four frameworks at a time).

src/docs/patterns/appbar.mdx
joebochill: It might be simplest to just include the material dimensions instead of having to go through the path of inheritance.

src/docs/patterns/appbar.mdx
joebochill: I don't mind having the link in multiple places, but should we use the same link (card) style...or something similar to reinforce the idea of "Hey this goes to the Material documentation"?

src/docs/patterns/appbar.mdx
huayunh: Added to the backlog as PXBLUE-1513 for the next grooming

src/docs/patterns/appbar.mdx
huayunh: Changed to the callout card. Their deep link is messed up -- it should have taken people directly to the #Specs section

src/docs/patterns/forms.mdx
huayunh: I like duplication better in this case. Our persona Erice will just scroll all the way to the bottom for design specs instructions; if he can't find any, he should not be forced to scroll all the way up for that callout card, or go googling for appbar: top pattern himself.

src/docs/patterns/appbar.mdx
joebochill: Did you push that change? I'm still seeing the same link.

src/docs/patterns/forms.mdx
joebochill: I agree, I'm cool with having it in two places

src/docs/patterns/appbar.mdx
```diff
@@ -35,3 +37,37 @@ In some situations, you may want to present more information in an app bar, such
...
+
+## Mobile
+
+<ImageGrid images={[Specs1Image]} regularWidth />
+
+## Desktop and Tablet
+
+If certain dimensions are not specified, refer to the dimensions suggested in the mobile section above.
+
+<ImageGrid images={[Specs2Image]} regularWidth />
```
joebochill, line 55: When it's implemented in code, that icon is still going to be on a 24px background:
image

Specifying it as 8 pixels might be confusing.

src/docs/patterns/appbar.mdx
joebochill: Did you push all of your local changes...this still looks the same

src/docs/patterns/appbar.mdx
huayunh: Interesting... in all the design files it is labeled 8px wide. Is it impossible to achieve this on the dev side?

joebochill: Not without a lot of hacking. The SVG is defined in a 24x24 viewbox, so all of the Paths are relative to that. You would have to actually go in and modify the paths somehow. This is on the React side...i'm not 100% sure about angular, but I suspect they're using the same viewbox for their icons.

huayunh: I got you... I should probably shy away from this one in particular then. It is funny that all the design sources (our Figma files, MD icon downloads, MD design specs) are doing the 8px wide trick, yet all our dev code is using a 24*24 viewbox

image

huayunh:
What the heck I swear I pushed

PR #227, opened 2020-07-27
Update angular design pattern links to point to monorepo

daileytj:
Fixes PXBLUE 1476.

- Update angular design pattern links to point to monorepo


jeffvg:
will stackblitz support loading up a specific page like we do for react code sandbox?

daileytj:
> will stackblitz support loading up a specific page like we do for react code sandbox?

Unfortunately at this time it doesn't. Code Sandbox supports this but we can't get it working there right now.

joebochill:
Do we also want to point side-navigation and icon-navigation to the component library for Drawer?

Might also be nice for these (Empty State, Side Navigation, Icon Navigation, maybe Login) to add a `// TODO` comment to switch these over to the "Component Recommendations style" instead of a "Interactive Code example" once we have one designed.

PR #230, opened 2020-08-05
Update react theme link

joebochill:
- Update the link for the react theme demo

## NOTE
The master branch equivalent for this should be merged in quickly after the themes change goes into the master branch (dev branch PR is here https://github.com/pxblue/themes/pull/96) because the folder structure/location of the demo will change


joebochill:
> why is this an embeded link?

Looks like it never changed from when the example was embedded. It doesn't make a big difference, but I'll update it.

PR #240, opened 2020-09-02
Tweak Form Pattern

joebochill:
For PXBLUE-1479.

- Re-order some content
- Add some new content
- Reword a few things
- Update spec images
- Updated content images (except for gifs)


emclaug2 (approved):

jeffvg (approved):
looks good

huayunh:
Not planning to update the other illustrations to abide by the design specifications?

And no illustrations for those transparent helper text?

joebochill:
> Not planning to update the other illustrations to abide by the design specifications?
> 
> And no illustrations for those transparent helper text?

Can you be more specific about which images are not following the spec? They look correct to me, except for the gif of the login screen.

And as far as the 'transparent' helper text, I thought we had agreed to "assume there is helper text" on the login screen only and others forms would shift their layout when errors appear (like in the gif). I can add another spec image that shows the space staying at 24/32 when there is no assistive text if that would be helpful.

huayunh:
Incorrect (should all be problems around the spacing between different elements)
image
(as well as the other two)

image

image

image

huayunh:
> I thought we had agreed to "assume there is helper text" on the login screen only and others forms would shift their layout when errors appear (like in the gif)

Not sure about the login screen ONLY part... I'd say check with other designers on this. 

One thing I realized is the helper text might wrap on a smaller screen and not wrap on a larger screen, thus resulting in different margins between two text fields. Take that into consideration too.

joebochill:
> One thing I realized is the helper text might wrap on a smaller screen and not wrap on a larger screen, thus resulting in different margins between two text fields. Take that into consideration too.

Sounds like a good enough reason to use a fixed spacing between fields (and not try to make assumptions about helper text): fixed space from bottom of helper text (if it exists) to top of next field or fixed space from bottom of first field (if no helper text) to top of next field. Login being a special case (if we want it to be) to prevent the fields from moving on error.

👍  👎  ?

I see what you're talking about with the images (I was only looking at the outer margins/padding). I will loop back around and update those in a separate PR if the rest of this content looks okay.

huayunh:
I called for a meeting on Friday afternoon, during which we can prob get some consensus on this.

huayunh:
Note: maybe also ask if the submit button should always be pinned at screen corners (top right corner or bottom right corner)

joebochill:
I updated all of our usage of slash to have a space on both sides (based on comment on other PR #242). If we don't want to do this it will be easier to do find replace on these than the other way (every url has multiple [character]/[character] instances....10000+ instances of that).

joebochill:
All of the images should be updated now, so if the content seems okay to us, I'd like to get it in dev so Kyle and Ian can have a look at the content and give any thoughts for follow-up changes.

src/docs/patterns/forms.mdx
```diff
@@ -160,54 +174,36 @@ For screens such as email verification where there is a single field with a know
...
     <ImageGrid
         images={[SubmitButtonOnBlurImage]}
         regularWidth
         caption={'The submit button is disabled until the user has entered something for all the required fields.'}
     />
+    <ImageGrid
+        images={[dismissKeyboardErrorAutoscrollImage, dismissKeyboardErrorTextImage]}
+        caption={[
+            'Left: The UI should auto-scroll to and focus on the first invalid field.',
+            'Right: On login screens, form fields should not shift when errors appear.',
```
huayunh, line 198: Just a note that these ImageGrids get displayed in a stacked way on mobile, and it would make no sense to call them "left" and "right"

src/docs/patterns/forms.mdx
```diff
@@ -218,14 +214,11 @@ For additional instructions, refer to [Material Design's design specs on "Text f
 If certain dimensions are not specified, refer to the dimensions suggested in the mobile section above.
 
 <ImageGrid images={[Specs4Image, Specs5Image]} regularWidth />
```
huayunh, line 216: the 40/32 margin is redundant here, since it is the same as the mobile dimension

huayunh:
Otherwise, go ahead and merge this PR 👏 

I tried my best to break the 71-comment record, I guess it is not gonna work this way

PR #248, opened 2020-09-10
Feature/app bar pattern

joebochill:
For PXBLUE-1477

- Update app bar content and images

image


src/docs/patterns/appbar.mdx
```diff
@@ -1,43 +1,86 @@
...
+        repository={'Dropdown Toolbar'}
+        angular={
+            'https://pxblue-components.github.io/angular/?path=/story/components-dropdown-toolbar--with-full-config'
+        }
+        react={'https://pxblue-components.github.io/react/?path=/story/components-dropdown-toolbar--with-full-config'}
+        ionic
+        reactNative
+    />
+    <Divider />
+</React.Fragment>
```
huayunh, line 55: I think all the other patterns are following the image-demo-text order? Correct me if I'm wrong

src/docs/patterns/appbar.mdx
```diff
@@ -1,43 +1,86 @@
...
+
+# Collapsible App Bar
 
 <ImageGrid
     images={[Image1, Image2]}
     caption={'A collapsible app bar shrinks and gets pinned to the top when the page is scrolled.'}
 />
```
huayunh, line 62: Better done with GIF

huayunh: Might also want to double-check the list item spacing

src/docs/patterns/appbar.mdx
```diff
@@ -1,43 +1,86 @@
...
 
-In some situations, you may want to present more information in an app bar, such as a large graphic, buttons, or multiple lines of text. In these cases, you can use a taller app bar that will shrink as the user scrolls down the page and be replaced by a basic app bar. This allows you the flexibility to present an eye-catching banner without permanently reducing the amount of usable space on the page. This is a pattern most commonly used for mobile applications, but could also be used at a larger scale.
+In some situations, you may want to present more than just a title and subtitle in an app bar, such as a large graphic, buttons, or multiple lines of text. In these cases, you can use a taller app bar that will shrink as the user scrolls down the page until it becomes a basic app bar. This allows you the flexibility to present an eye-catching banner without permanently reducing the amount of usable space on the page. This is a pattern most commonly used for mobile applications, but could also be used for web.
```
huayunh, line 65: Mention the word "collapsible" at least once in the introduction paragraph

> In these cases, you can use a taller app bar that will shrink as the user scrolls down the page until it **collapses** into a basic app bar.

src/docs/patterns/appbar.mdx
```diff
@@ -1,43 +1,86 @@
...
 
-## Search Bar
+The collapsible appbar should transition smoothly between the expanded and collapsed states — this means you will need to account for how the app bar will look throughout the transition. Because elements will disappear as the app bar shrinks, you'll need to use dynamic text scaling, opacity transitions, and / or other techniques to do this smoothly, depending on your content.
```
huayunh, line 67: We shall probably be more opinionated on this. Ideally, it should look like the Header component in React Native Component Library, but at the same time, I understand that this pattern can be a bit tricky in implementing the animation.

src/docs/patterns/appbar.mdx
huayunh: Maybe screen record the React Native component showcase?

src/docs/patterns/appbar.mdx
```diff
@@ -1,43 +1,86 @@
...
+
+<Divider />
+
+# Search Bar
 
 <ImageGrid
     images={[Image3, Image4]}
```
huayunh, line 74: This one probably also needs to be revisited on the list item spacing

huayunh: And better done with Gif too, to show the transition animation

src/docs/patterns/appbar.mdx
```diff
@@ -1,43 +1,86 @@
...
 
-A very common pattern seen in many types of applications is the ability to search for data on a page. This is achieved by placing a search icon/button in the app bar. When this button is clicked, a white Search Bar slides in from the right to cover the default toolbar. As the user enters text into the search field, the data is filtered in real-time.
+A very common pattern seen in many types of applications is the ability to search for data on a page. This is achieved by placing a search icon/button in the app bar. When this button is clicked, a white search bar slides in from the right to cover the default app bar. As the user enters text into the search field, the data is filtered in real-time. A back icon will close the Search Bar and clear the search results.
+
+## Auto-complete
+
+For more advanced searching, you may wish to include an auto-complete feature. You can implement autocomplete by including a contextual menu below the search bar with suggested search items. Alternatively, you can use type-ahead to automatically update the recommended search query in the search field as the user types (Google search does this).
```
huayunh, line 83: Would like to see two illustrations for the two auto-complete variants mentioned here

src/docs/patterns/appbar.mdx
```diff
@@ -1,43 +1,86 @@
...
 
 <Divider />
```
huayunh, line 85: One thing I was wondering — why are ng material and MUI both using 64px as the app bar height? I don't think the 64px is mentioned anywhere in Material's design spec.

src/docs/patterns/appbar.mdx
```diff
@@ -1,43 +1,86 @@
...
 />
 <DemoCard repository={'search-bar'} angular={'https://stackblitz.com/github/pxblue/angular-design-patterns?file=src%2Fapp%2Fpages%2Fapp-bar%2Fsearch-bar%2Fsearch-bar.component.ts'} react={'https://codesandbox.io/s/github/pxblue/react-design-patterns/tree/master?file=/src/pages/app-bar/search-bar/index.tsx&initialpath=search'} ionic reactNative />
 
-A very common pattern seen in many types of applications is the ability to search for data on a page. This is achieved by placing a search icon/button in the app bar. When this button is clicked, a white Search Bar slides in from the right to cover the default toolbar. As the user enters text into the search field, the data is filtered in real-time.
+A very common pattern seen in many types of applications is the ability to search for data on a page. This is achieved by placing a search icon/button in the app bar. When this button is clicked, a white search bar slides in from the right to cover the default app bar. As the user enters text into the search field, the data is filtered in real-time. A back icon will close the Search Bar and clear the search results.
```
huayunh, line 79: I think not all the applications do a mega white search bar like our docit. A lot of them do it in a small text field, like what we have on the iconography page icon browser. 

image

src/docs/patterns/appbar.mdx
joebochill: It depends on the content too. Are we telling people what to put inside of it (i.e, a collapsible app bar has three lines of text and a background image) or are we leaving that up to them and saying the pattern is just "big toolbar becomes small and make the content transition cleanly" and they could put anything they want in it?...I was assuming the latter.

src/docs/patterns/appbar.mdx
joebochill: Good question...i don't know. It's possible it was part of their specs before and got removed and the component libraries just carried on.

src/docs/patterns/appbar.mdx
joebochill: Do we want to include the mini one as another option? What are the rules for using one versus the other?

src/docs/patterns/appbar.mdx
joebochill: We don't actually have an implementation that does this properly yet. The react native one works on click, not scroll. I built one in RN for another project that we could potentially use. I definitely don't want to try to get Figma to animate that properly.

src/docs/patterns/appbar.mdx
huayunh: Hmmm, so should we shy away from the desktop dimension then?

huayunh: And can we have an img for collapsible app bar?

src/docs/patterns/appbar.mdx
huayunh: I am more inclined to yes, but I’d say have some design meetings around it.

src/docs/patterns/appbar.mdx
huayunh: I see what you mean here. I personally think we shall encourage the use of animations like RN Header bar by including a pic of it, then we can mention that there are several other different ways you can animate it

huayunh:
Developer section should mention drop down tool bar

src/docs/patterns/appbar.mdx
joebochill: As far as I can tell, there's no consistency.

I see on different pages:
`image-text-demo-image`
`image-text-demo`
`image-demo-text`
`text-image-demo`
`text-demo`
and
`demo-text`

I'm kind of preferential to text first to give some context around the images/demos.

joebochill: the ones currently doing image-demo-text are all ones that haven't been overhauled yet

src/docs/patterns/appbar.mdx
```diff
@@ -1,12 +1,10 @@
 import { DemoCard, ImageGrid, MaterialDesignDescription, Divider } from '../../app/components';
 import Banner from '../../app/assets/design-patterns/app-bar/appbar-banner.png';
-import Image1 from '../../app/assets/design-patterns/app-bar/collapsible-appbar/collapsible-appbar-expanded.png';
-import Image2 from '../../app/assets/design-patterns/app-bar/collapsible-appbar/collapsible-appbar-collapsed.png';
 import Collapsible from '../../app/assets/design-patterns/app-bar/collapsible-appbar/collapsible.gif';
+import Search from '../../app/assets/design-patterns/app-bar/search-bar/search.gif';
```
huayunh, line 4: (1) I think the search bar should slide in and cover the app bar, rather than pushing it away
(2) I realized that doc-it's search bar does not match exactly with this screenshot... We might need to update one of it

src/docs/patterns/appbar.mdx
joebochill: I wouldn't shy away from it, unless we're planning on overhauling our themes to override the default behaviors. I don't see a reason to at this point in time. I think the 64px works better when you have a subtitle (56 feels cramped)...material doesn't do subtitles in toolbars as far as I can tell.

src/docs/patterns/appbar.mdx
joebochill: Updating doc-it will be the faster way since it's already implemented like this in the RN component library

joebochill: ...layout wise. It doesn't do the slide.

joebochill: We can tell people it's a slide that occurs in 0ms

src/docs/patterns/appbar.mdx
huayunh: I agree on the 64px works better with the subtitle part. Just be aware that DSM and all other design files are using 56px

src/docs/patterns/appbar.mdx
```diff
@@ -1,45 +1,133 @@
...
+    <Divider />
+</React.Fragment>
+
+# Search Bar
 
 <ImageGrid
-    images={[Image3, Image4]}
+    regularWidth
+    images={[Search]}
     caption={'When the search icon is clicked, a white search bar will slide in and cover the default tool bar. '}
```
huayunh, line 90: We've got two actionable icons here, both indicate some dismissal actions. Here's what I suggest for the animation:
search-bar

src/docs/patterns/appbar.mdx
```diff
@@ -1,45 +1,133 @@
...
 
-A very common pattern seen in many types of applications is the ability to search for data on a page. This is achieved by placing a search icon/button in the app bar. When this button is clicked, a white Search Bar slides in from the right to cover the default toolbar. As the user enters text into the search field, the data is filtered in real-time.
-
-<Divider />
+A very common pattern seen in many types of applications is the ability to search for data on a page. This is achieved by placing a search icon/button in the app bar. When this button is clicked, a white search bar slides in from the right to cover the default app bar. As the user enters text into the search field, the data is filtered in real-time. A back icon will close the Search Bar and clear the search results.
+
+If your search is not a global-level search, you can place a smaller search input near the content being searched (e.g., a card header).
+
+<React.Fragment>
+    <ImageGrid
```
huayunh, line 98: Would prefer this to be more hi res

src/docs/patterns/appbar.mdx
```diff
@@ -1,45 +1,133 @@
...
+</React.Fragment>
+
+## Auto-complete
+
+For more advanced searching, you may wish to include an auto-complete feature. You can implement autocomplete by including a contextual menu below the search bar with suggested search items. Alternatively, you can use type-ahead to automatically update the recommended search query in the search field as the user types (Google search does this).
+
+<React.Fragment>
+    <ImageGrid
+        regularWidth
+        images={[AutoVariants]}
```
huayunh, line 123: Don't like the elevation; see figma for a potential fix (or you can do what docit is doing)

huayunh:
I don't know about this image. I think mobile should look the same as the desktop (so that the user doesn't have to move their fingers up and down) -- ask other designers about this

And there are too many dividers in pic 1, see Material Design's guideline: https://material.io/components/dividers

image

src/docs/patterns/appbar.mdx
huayunh: Source file in Figma, if you want to make changes or remove that bottom stroke there

joebochill:
> And there are too many dividers in pic 1, see Material Design's guideline: https://material.io/components/dividers

Seems like Material has somewhat contradictory advice. They use dividers to separate list items above and then say not to do it:
image
image

I'm fine with taking them out in this case though.

joebochill:
> I don't know about this image. I think mobile should look the same as the desktop (so that the user doesn't have to move their fingers up and down) -- ask other designers about this

Up until this point, we've been avoiding the use of any sort of dropdown menus on mobile in favor of bottomsheets. I'm assuming that this is to more closely align with the underlying native behaviors of select pickers.

huayunh:
> Seems like Material has somewhat contradictory advice. 

My understanding is that count as "grouped items" :P 

But definitely not one divider per single line

huayunh (approved):
> Up until this point, we've been avoiding the use of any sort of dropdown menus on mobile in favor of bottomsheets. I'm assuming that this is to more closely align with the underlying native behaviors of select pickers.

Let's ask other designers about it; I played around with an Android phone and it seems they are avoiding this. Our dropdown toolbar component also does not behave this way by default without any hacks.

PR #255, opened 2020-09-18
Emphasize more on the formats

huayunh:
Incorporate some Ian feedback & expand 

- Added a section "International Contacts"
- Added a "commonly influenced formats" list
- Added a link to Wikipedia

#### Screenshots:
image


src/docs/patterns/internationalization.mdx
```diff
@@ -52,7 +53,22 @@ When showing a list of languages, the language names should be presented in thei
...
+Units, currency, and formats are usually linked with a particular locale. However, there can be differences in usage even within a single locale. You should provide users with the option to change these values (if appropriate for your application) in their locale settings.
+
+Some commonly influenced formats include:
+
+-   Phone number
+-   Names (given name first or last)
+-   Address (street name first or country name first)
+-   Date and time
+-   Numbers (comma vs. period)
+-   Punctuations
```
joebochill, line 65:
```suggestion
-   Phone number
-   Names (e.g., given name vs. surname order)
-   Address (e.g., street name first or country name first)
-   Date and time
-   Numbers (comma vs. period)
-   Punctuation
```
any examples of where punctuation differs? I know quotations are different, but was there something else you had in mind?

src/docs/patterns/internationalization.mdx
```diff
@@ -52,7 +53,22 @@ When showing a list of languages, the language names should be presented in thei
...
+
+## International Contacts
+
+<ImageGrid images={[FormsImage]} caption={'Allow the users to put in an international phone number.'} regularWidth />
+
+Your users may register a new account or create a new contact for anyone around the globe. You should provide options for them to register with a local phone number.
```
joebochill, line 71: I would skip this as a section on its own, particularly since phone number is an optional field for registration. I would just move the image to the previous section and give it a descriptive caption like "Consider your global audience when crafting forms — you may need to support multiple input formats such as for a phone number."

src/docs/patterns/internationalization.mdx
huayunh: Mostly for that quotation thing«»  but… there can be a lot of Asian things that differ from those Latin based languages

。、《》「」

src/docs/patterns/internationalization.mdx
huayunh: Yeah! This is certainly better. I tried merging the two but couldn’t find a proper way. Thanks.

huayunh:
Not gonna run yarn indexer this time... We can run that for the final push of dev->master

joebochill (approved):

PR #268, opened 2020-09-24
Feature/ionic links

emclaug2:
- Add ionic links
- Remove react-native and ionic links from empty-state pattern


src/docs/patterns/appbar.mdx
```diff
@@ -47,8 +47,6 @@ The basic app bar provides information about the current screen. Typically, this
             'https://pxblue-components.github.io/angular/?path=/story/components-dropdown-toolbar--with-full-config'
         }
         react={'https://pxblue-components.github.io/react/?path=/story/components-dropdown-toolbar--with-full-config'}
-        ionic
```
joebochill, line 50: You could leave ionic here and point it to Angular if you want

joebochill (commented):
Should we be removing / commenting out all of the RN examples for now since the import into Snack is not working?

emclaug2:
> Should we be removing / commenting out all of the RN examples for now since the import into Snack is not working?

If our interactive examples are known to not work at this moment I'll comment them out.

src/docs/patterns/appbar.mdx
emclaug2: I think it's best to avoid pointing to the same resource.

emclaug2:
All the RN interactive demo links are removed right now.

src/app/components/patterns/DemoCard.tsx
```diff
@@ -127,13 +127,13 @@ export const DemoCard: React.FC<DemoCardProps> = (props): JSX.Element => {
                 {ionic && (
                     <DemoButton repository={repository} framework={'ionic'} url={ionic === true ? undefined : ionic} />
                 )}
-                {reactNative && (
+                {/*reactNative && (
```
joebochill, line 130: Only the react-native-design-patterns are not working. I'm not sure if we link out to anything else (component docs?) from a DemoCard, but you might want to double check that.

emclaug2: I looked around and didn't see any react-native resources that have demos, this fix should be good to go.

joebochill (approved):

PR #281, opened 2020-10-03
Use clsx; fix double selection error

### LINKED ISSUE #276: Iconography page selects both apple icons
#### Describe the bug

The iconography page selects all icons under the same name when there are dup names

It should select icons with only the correct family

1. Go to https://px-blue.github.io/style/iconography?icon=Apple
2. Observe that Material-UI's Apple and PX Blue's Apple both get selected.

#### Screenshots or links to minimum reproduction example
image

#### Environment

		
-- Saint Joseph

huayunh:
Fixes #276

Use `clsx` because I saw this in the source code:
image

### DIFF
--- a/src/app/components/icons/IconBrowser.tsx
+++ b/src/app/components/icons/IconBrowser.tsx
@@ -36,7 +36,7 @@ type LetterGrouping = {
 };
 
 const hideResultsThreshold = 20;
-const Icons: MatIconList = MuiIcons;
+const PXBlueIcons: MatIconList = MuiIcons;
 const MaterialIcons: MatIconList = AllMaterialIcons;
 
 const useStyles = makeStyles((theme: Theme) => ({
@@ -133,7 +133,7 @@ const createIconList = (): Icon[] => {
     const iconList: Icon[] = [];
     getFilteredIcons().forEach((icon: DetailedIcon) => {
         const mui = getMuiIconName(icon.filename);
-        if (Icons[mui]) {
+        if (PXBlueIcons[mui]) {
             iconList.push({ name: icon.filename.replace(/\.svg/, ''), isMaterial: false });
         }
     });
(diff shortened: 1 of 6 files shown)


src/app/components/icons/IconCard.tsx
```diff
@@ -39,7 +40,7 @@ export const IconCard: React.FC<IconCardProps> = (props): JSX.Element => {
 
     const { component: Component, name, showLabel, style, selected, iconSize, className } = props;
     return (
-        <div className={`${classes.wrapper} ${selected ? classes.selected : ''} + ${className}`} style={style}>
+        <div className={clsx(classes.wrapper, { [classes.selected]: selected }, className)} style={style}>
```
joebochill, line 43: I actually saw a couple instances where we were using the `+` inside of a template literal. You may want to do a quick sweep and see if you can find any more.

src/app/components/icons/IconBrowser.tsx
```diff
@@ -333,7 +333,8 @@ export const IconBrowser: React.FC = (): JSX.Element => {
                                                             selected={
                                                                 focusedIcon &&
                                                                 getMuiIconName(focusedIcon.name) ===
-                                                                    getMuiIconName(icon.name)
+                                                                    getMuiIconName(icon.name) &&
```
joebochill, line 336: You probably also need to add a param to the query string for which icon class (material or blue) the icon is. Otherwise, we can never deep link to the MUI apple icon.

joebochill: We can default to either one if none is specified.

huayunh: My implementation: 

`icon=Apple&isMaterial=` | `true` | `false` | `asdf` | _none_
-|-|-|-|-
PX Blue Apple |                 | selected | selected | selected
Material Apple | selected |                |                 | 

`icon=Apps&isMaterial=` | `true` | `false` | `asdf` | _none_
-|-|-|-|-
Material Apps | selected |                |                 | selected

joebochill (approved):

PR #282, opened 2020-10-05
fix security issues

jeffvg:
- added yarn-audit-fix package
- https://www.npmjs.com/package/yarn-audit-fix
- 

#### Screenshots:
- old
4338 vulnerabilities found 
Severity: 4307 Low | 19 Moderate | 12 High

- new
0 vulnerabilities found - Packages audited: 2307

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
         "react-ga": "^2.7.0",
         "react-redux": "^7.2.0",
         "react-router-dom": "^5.1.2",
-        "redux": "^4.0.5"
+        "redux": "^4.0.5",
+        "yarn-audit-fix": "^3.2.6"
     },
     "devDependencies": {
         "@babel/core": "7.7.4",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -43,7 +43,8 @@
         "react-ga": "^2.7.0",
         "react-redux": "^7.2.0",
         "react-router-dom": "^5.1.2",
-        "redux": "^4.0.5"
+        "redux": "^4.0.5",
+        "yarn-audit-fix": "^3.2.6"
```
joebochill, line 47: For future, I would make this a devDependency instead of a normal dependency. It doesn't make a difference for doc-it (nobody is installing doc-it in their project), but it will if we want to use this (as I suspect we probably will) in other repositories like the component libraries.

joebochill (approved):

package.json
jeffvg: agree but the 'yarn add --d yarn-audit-fix' wasn't working so I used the 'yarn add yarn-audit-fix -D' instead and didn't realize. Strange that it's in dependencies and not devDependencies

joebochill: I do
```
yarn add --dev <package-name>
```
for installing dev dependencies. According to the docs, the shorthand is `-D`
https://classic.yarnpkg.com/en/docs/cli/add/#toc-yarn-add-dev-d

joebochill: it's `--dev` or `-D`, `--d` isn't valid yarn install syntax. 

Everything gets installed to the same node_modules folder though...you can just cut/paste that line in the package.json file 😉

PR #285, opened 2020-10-07
Add "contextual search"

huayunh:
Just adding in a few thoughts while talking to an adopter.

- Elaborate on "contextual search"
- design spec "600px" -> "600dp"
- I feel like we should mention the use of a secondary toolbar somewhere, and link out to the navigation pattern (once Kyle's PR is finished). But I couldn't find out a proper place to insert that comment.

## Screenshots
image


huayunh:
^ markdown link thing is fixed in #284

src/docs/patterns/appbar.mdx
```diff
@@ -107,27 +110,30 @@ The collapsible appbar should transition smoothly between the expanded and colla
...
+    images={[AutoVariants]}
+    caption={[
+        '1. Search suggestions are provided in a dropdown menu.',
+        '2. The most likely search is shown in the search field as the user types.',
+    ]}
+/>
+
+## Contextual search
+
+If your search is not a global-level search, you can place your search input in different UI hierarchies to indicate the search context.
```
joebochill, line 128:
```suggestion
If your search is not a global-level search, you can relocate the search input to indicate the context of the search.
```

src/docs/patterns/appbar.mdx
```diff
@@ -107,27 +110,30 @@ The collapsible appbar should transition smoothly between the expanded and colla
...
 <React.Fragment>
     <ImageGrid
-        regularWidth
-        images={[AutoVariants]}
+        images={[SiteSearch, PageSearch, ComponentSearch]}
         caption={[
-            '1. Search suggestions are provided in a dropdown menu.',
-            '2. The most likely search is shown in the search field as the user types.',
+            '1: Site-wide search, a.k.a. global search, is always placed in the primary app bar.',
+            '2: Page-wide search, usually placed in a secondary toolbar.',
```
joebochill, line 135: The second one feels strange to me. Are we steering people to do double appbars like that? What if you have a sitewide search and a page-level search? What about for Mobile?

Might want to call out some of these cases.

huayunh: It was more like an effort to distinguish actions at different hierarchies. So if you place something in a primary toolbar, it is assumed that this will be a global action. If you place icons in a secondary toolbar, then it applies to that page only. If you put it on a component bar, then it pertains to that component.

This will be better suited to a behavior section on its own. But that requires some extra effort so I did not do that...

joebochill: I think secondary toolbar for search just feels weird to me. I would assume that global or page-level search would go in the main toolbar and it would just depend on context which one it was (I wouldn't expect many applications to a global search **and** page-level search at the same time). Adding a secondary toolbar just to do page-level search seems overkill when you have a perfectly good primary toolbar. Other patterns (appbar) describe using the main toolbar for page-level actions as well.

joebochill:
@huayunh do we need to have a design discussion around the use of the secondary toolbar?

huayunh:
We can sit down and chat on this

joebochill:
@huayunh can you ask @Horchen154 about and get his thoughts since he wasn't at the chat last week?

I did see the Amazon app uses a search in the secondary toolbar to search your past orders, while still having a global product search in the primary toolbar. 
IMG_3169

I wouldn't consider this a good example (I always type in the wrong one and get annoyed), but it is an example.

joebochill (approved):

PR #291, opened 2020-10-20
Feature/halloween theme

huayunh:
Boo.

# Changes proposed in this Pull Request
- Added a spooky theme that is currently scheduled to last from 10/25 to 11/02
- The theme is designed to match Kyle's dark theme design as close as possible

To test, change line 36 of `doc-it/src/__configuration__/themes/schedule.ts` to `start: new Date(0, 8, 25),`, so that the starting date will become September 25th instead of October.

# Screenshots

(No copyright worries, everything is home-made or copyleft)

image

image

image

# Known Issues

Listed below are some known issues. They might be discussed during office hours.

### 1
image
Chevron is not changing color. I don't know how to make it orange[500]... might have to change the drawer API.

### 2 
This `<code>` piece is not changing color as specified in the config file
image

Same with this `<pre>` block inside `IconMenu`
image

### 3
Would be good if the selected tab font color is also orange 500
image

### 4
The loading state of the resources/roadmap pages is still using the light theme. Not sure how to change the sass through js.


emclaug2:
This looks very professional; very impressive Halloween theme & really great landing page image.  I'm going to sift through the code in a little bit but visually I wouldn't change anything.

emclaug2:
I just caught one thing, since the Halloween theme is a dark theme, can you check the loading states on the Roadmap page?  Currently it's a light-themed loading state, so the page flashes white before the data loads in.

joebochill:
## 1
The chevron is inheriting the color from the wrapper div:
image
Should be able to override this by changing the `activeItemIconColor` prop

## 2
There is another item that you need to add to the markdownMapping: inlineCode (it can be exactly the same as code)

Unfortunately the icon bottomsheet is not markdown so the markdown mapping config will not apply to it. If you want to make it work, you will probably need to update index.css and conditionally add a top-level 'theme' class (e.g., 'halloween') based on the current active theme and then do something like:
```css
.halloween pre, .halloween code{
  //special color overrides
}
```

## 3
image


## 4
You could probably do something similar to number 2 part 2, or just leave it for now.

joebochill:
This looks sick by the way 🔥

joebochill:
Also, I would suggest running the halloween theme for the whole month of October (1-31)

src/app/pages/MarkdownPage.tsx
```diff
@@ -16,7 +17,16 @@ export type MarkdownPageProps = {
...
+    const theme = useTheme();
+    if (props.background === 'light') {
+        useBackgroundColor(
+            theme.palette.type === 'light' ? theme.palette.background.paper : theme.palette.background.default
+        );
+    } else if (props.background === 'dark') {
+        useBackgroundColor(theme.palette.type === 'light' ? Colors.white[200] : theme.palette.background.default);
+    } else {
+        useBackgroundColor(props.background);
+    }
```
joebochill, line 29: I think this violates the rules of hooks (having a hook call in a conditional block). You may want to refactor this slightly to do the conditional logic to get the colors and then have a single call to useBackgroundColor after that.

joebochill (approved):

jeffvg (approved):
This really looks awesome....

huayunh:
Let's go :)

PR #296, opened 2020-11-07
Updated Iconography Page

joebochill:
For PXBLUE-1530

- Update iconography page to match Figma design
- Implement Persistent/Closable Drawer
- Refactor and Reorganize everything
- Performance optimizations
- Implement icon categories and filtering/searching by name and tags
- Make sure all of the deep linking still works

#### Screenshots:
image


emclaug2:
This is really great Joe!  I'm looking at the code to understand how some of the optimizations work; I'm curious how you got the history updates to work without the sluggishness of before.

src/app/components/icons/IconSearchBar.tsx
```diff
@@ -0,0 +1,128 @@
...
+// import * as Colors from '@pxblue/colors';
+import { titleCase } from '../../shared';
+
+const useStyles = makeStyles(() => ({
+    searchBar: {
+        display: 'flex',
+        flexWrap: 'wrap',
+    },
+    searchField: {
+        // flex: '1 1 0px',
```
emclaug2, line 30: Can this class be removed?

src/app/components/icons/IconSearchBar.tsx
```diff
@@ -0,0 +1,128 @@
...
+        // minWidth: 300,
+        // marginRight: theme.spacing(2)
+        // borderRadius: theme.shape.borderRadius,
+        // backgroundColor: Colors.white[50],
+        // color: theme.palette.text.primary,
+        // marginTop: theme.spacing(1),
+        // marginRight: theme.spacing(3),
+    },
+    categoryPicker: {
```
emclaug2, line 39: And this one too?

src/app/components/icons/IconSearchBar.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+type SearchBarProps = HTMLAttributes<HTMLDivElement> & {
+    onSearchChange: (event: ChangeEvent<HTMLInputElement>) => void;
+    onCategoriesChanged: (event: ChangeEvent<{ name?: string; value: any }>) => void;
+    iconCategories: string[];
+};
+export const IconSearchBar: React.FC<SearchBarProps> = (props): JSX.Element => {
+    const { onSearchChange, onCategoriesChanged, iconCategories, ...divProps } = props;
+    const classes = useStyles();
+    const { iconSearch = '' } = useQueryString();
```
emclaug2, line 53: Is this something within the scope of this PR? : 
When a search happens, should the URL be updated to include the search query string?
e.g. "http://localhost:3000/style/iconography?iconSearch=tree"

src/app/components/icons/IconSearchBar.tsx
```diff
@@ -0,0 +1,128 @@
...
+                <FormControl className={classes.categoryPicker} variant={'outlined'} fullWidth>
+                    <InputLabel id="category-select-label">Icon Categories</InputLabel>
+                    <Select
+                        labelId="category-select-label"
+                        label={'Filter By Category'}
+                        fullWidth
+                        multiple
+                        value={selectedCategories}
+                        onChange={handleFilterChange}
+                        // input={<OutlinedInput />}
```
emclaug2, line 108: Remove?

src/app/components/icons/IconSearchBar.tsx
```diff
@@ -0,0 +1,128 @@
...
+                        endAdornment: (
+                            <InputAdornment position={'end'}>
+                                <Search />
+                            </InputAdornment>
+                        ),
+                    }}
+                />
+            </Grid>
+            <Grid item xs={categorySize}>
+                <FormControl className={classes.categoryPicker} variant={'outlined'} fullWidth>
```
emclaug2, line 99: Should this filter have an icon `startAdornment`?

emclaug2 (changes requested):
This is a fantastic refactor; thank you Joe.  

I've added a couple comments and have noticed 2 small things:

1.  The Eaton footer width does not adjust when sideBar visible:

image


2. The RN dev accordion has a double divider when opened:

image

src/app/components/icons/IconSearchBar.tsx
joebochill: I wondered the same thing. I checked the existing implementation and it doesn't do that, so I left it out for the time being.

src/app/components/icons/IconSearchBar.tsx
joebochill: Yes, I mean't to go through and do an unused class/var cleanup but forgot. I'll do that today.

src/app/components/icons/IconSearchBar.tsx
joebochill: It can. I tweaked the design a little from what was in the original mockup (it felt weird having two different styles of input and the 'group-work' icon didn't quite feel appropriate for categories. Open to suggestions on this (@huayunh @Horchen154 )

joebochill:
> This is really great Joe! I'm looking at the code to understand how some of the optimizations work; I'm curious how you got the history updates to work without the sluggishness of before.

To be honest, I'm still trying to unpack a lot of it myself. I was having the same issues you ran into with performance. Any time I wanted to add a new feature, suddenly the entire icon tree was rerendering and taking forever...whenever I could find a workaround for one thing, the next thing would break it again.

Ultimately, I ended up looking at how Materil-UI did their icon browser (https://material-ui.com/components/material-icons/) since it's super responsive and lists all of the icons just like we want to. I stole a bunch of their code and did a little refactoring to combine the two icon sets and do the categories.

I think the key performance improvement pieces are from:
- React.useMemo (not exactly sure how this works yet, but has something to do with limiting the number of redraws)
- using FlexSearch library to search (instead of doing our own manual looping implementation)

src/app/components/icons/IconSearchBar.tsx
huayunh: If you want to do multi-select, use a bunch of inline chips under the search bar. 

image

huayunh: Feels strange to do multi-select in a dropdown.

emclaug2: Something like this should have been discussed in the design stage of making this page, not implementation.  I think it's fine for now and if it should be changed that's a separate story.

huayunh:
This is def beyond my expectation. Fast and looking great!

## 1
Icons shall be horizontally aligned for a slightly better scanability
image

## 2
When I dismiss the side-sheet, there should be no icon selected. (I see that you changed the URL, but the icon is still selected)

## 3
Is there any way to trigger the download upon clicking the button? (Tried searching wild for this, looks like GitHub API doesn't allow this)

## 4
Would like to have a "copy icon name" icon next to the icon name on the side sheet 

## 5
There are a few Material Design icons listed under the "Obsolete Material Icons", such as 3D Rotation, 360, 4K, etc. They should be listed in their respective categories.

## 6
image
This code piece should use `3d_rotation` as the icon name instead. (same with a bunch of other icons with numbers in them)

## 7
You may want to do a temporary drawer on mobile

## 8
The icon spacing is _huge_ for `lg` screen size and below. 
image

## 9
image
The bottom border of the "React Native" panel seems extra wide when the panel is expanded.

src/app/components/icons/IconSearchBar.tsx
joebochill: I agree with Evan — if we want to change the design here, we should do it in a separate story. The design that we were working from uses a dropdown.

There are 30 categories right now - that might be a lot of chips to place under the search bar.

joebochill:
for 1) if we align the icons, we can't center the content. We would either have icons that sit higher in the selected area:
image
or we would have different sizes selection areas for different icons.

joebochill:
For 3) I spent a long time messing around with this and it looks like it is not supported by GitHub right now for the SVG file type.

joebochill:
For 4) what do you want this to look like?

joebochill:
For 5) These icons do not exist in the metadata from material (so we don't have any information about the category or tags). These icons are deprecated (no longer a part of material). I would lean towards hiding them entirely instead.

joebochill:
Nice catch on 6

joebochill:
For 8, I think we can get away with 4 across instead of three...it's a balancing act with making sure things aren't overlapping when the drawer is open. I went with a Grid implementation instead of a wrapping flex container as a simple solution to the "everything shifts down the page when you open the sidebar" problem rather than trying to do a 'scroll-to' type of thing.

huayunh:
For 1): Happy with what's in your screenshot
4):
image
5): 
I am talking about those material design icons with alphanumeric names, such as "360", "3d_rotations", etc. In fact, those two icons are the first two listed on that metadata list.

joebochill:
> Nice catch on 6

Fun fact, this is broken in master too...so it's been wrong for ~3 years. I should be able to fix it though.

src/app/components/icons/IconSearchBar.tsx
Horchen154: I would stay away from using chips instead of the dropdown. What would be the use case for selecting multiple options?

Horchen154:
Im not seeing anything that really sticks out that hasnt been addressed already. Is there any additional area of the design that is in question?

src/app/components/icons/IconSearchBar.tsx
joebochill: 🤷 greater flexibility than choosing a single category?

joebochill:
> Im not seeing anything that really sticks out that hasnt been addressed already. Is there any additional area of the design that is in question?

I think the only big question was around the categories / using chips or the dropdown. I think we're good for now (if we want to change those in the future, we can - let's see how it fares with users).

joebochill:
I think I have addressed all of the feedback at this point, with the exception of the copy icon name request...i'll do that quickly in the morning.

jeffvg:
Do we need any type of filtering on icon styles? (filled, outlined, rounded, two-tone ...etc)

Reason I asked is the DOM is pretty large with like 7k elements where the material site is slightly smaller, around 5k in elements on initial load. Although both are a hit on performance.

can we do this?

image

joebochill:
> Do we need any type of filtering on icon styles? (filled, outlined, rounded, two-tone ...etc)
> 
> Reason I asked is the DOM is pretty large with like 7k elements where the material site is slightly smaller, around 5k in elements on initial load. Although both are a hit on performance.
> 
> can we do this?
> 
> image

We could do that (I even precoded some of it to do that for the future), but I don't think it is going to save us any DOM elements. I already limit it to just the filled icons:

image

so none of the other styles are ever rendered at all. The extra DOM elements vs. MUI is likely coming from the extra ~200 PX Blue icons that we have (~15% more than base Material) plus the progress icons and the details sidebar (which MUI don't have).

There's also the issue of our PX Blue icons being a mix of styles (some are outlined, some are two tone, most are filled) and not having the different styles for all of them...we haven't organized these as well as material has, so it's unclear to me exactly how such a filter would apply to the PX Blue icons.

huayunh:
about that filled vs. outlined vs. sharp vs. two-tone... Eric mentioned to me once that they are not encouraging people to use the icons other than the filled version. 

The only exception is when you need to use icons to indicate some status change, like a filled version = on, outlined version = off / absent, etc. But 99% of this use case is covered in the existing icon set.

image

jeffvg:
did we intend to leave side drawer open when we nav away and back to icon page? just wondering

image

joebochill:
> did we intend to leave side drawer open when we nav away and back to icon page? just wondering

Hmm, probably not. Good catch. I'll have a look into that.

jeffvg:
are we able to display our site search while the side drawer is open?

image

jeffvg:
I'm having issues with browser emulator (iPad) and scrolling the icon side drawer when developer usage area is expanded. Really need to see it on real device. Other than that is looks great!

huayunh:
For those listed in the obsolete icons, take out the "tag / family" if N/A

take out the developer's instructions if not applicable. I think those obsolete icons only work for React Icon components?

image

huayunh:
And take out the download as svg / png too, they are not available through Google

huayunh:
image
Temporary drawer should not push content / push the app bar to the side

huayunh:
image
I suggest adding a few more pixels to the dropdown menu height to indicate menu items overflow

huayunh:
I suggest when there is only one category selected, change the "1 selected" to the category name
image

huayunh:
Between screen width of 1100 and 1260, the two input box are side-by-side at equal width

image

I prefer having the search bar in the "dominating" position, like what we have for them at the larger width

image

huayunh:
I suggest adding `edge={'start'}` to the `Checkbox` component on line 95 of `/src/app/components/icons/IconSearchBar.tsx`

huayunh:
Suggest making a new JIRA story to split this page into two: iconography guide (how to use) and icon resources (what icons do we have)

huayunh:
thumbs up otherwise.

I shall take more time to study your code structure here. Looks like a masterpiece.

joebochill:
Rather than conditionally hiding developer instructions, tags, and download buttons for obsolete icons, I have removed them entirely from the list. As mentioned earlier, we probably don't want to be encouraging people to use these at all in PX Blue projects, so best not to show them. If they are in the special group of people using React and ho want to use the motorcycle helmet icon, we can show them how to get it from MUI. But honestly, those obsolete icons are probably going to disappear from MUI anyway the next time they release an icons update because they are dynamically creating their list from the material metadata (plus the couple extra ones for github, etc.).

jeffvg (approved):
looks very nice all

huayunh:
Can we change the png download to 24dp instead of 18dp? 24 dp is a more common use case

Icons with alphanumeric names ("Three D Rotation", "Four K") cannot be downloaded. Should modify names to "3d_rotation", "4k", respectively

emclaug2 (approved):

PR #314, opened 2020-12-02
Add Hanukkah, Kwanzaa, and New Years themes

daileytj:
Fixes PXBLUE-1674.

- Add Hanukkah theme.
- Add Kwanzaa theme.
- Add New Years theme


jeffvg:
some of dates look off

jeffvg:
export const NewYearsSchedule: Schedule = {
    start: new Date(0, 11, 31), // Dec 31
    end: new Date(0, 0, 1), // Jan 1

daileytj:
> export const NewYearsSchedule: Schedule = {
> start: new Date(0, 11, 31), // Dec 31
> end: new Date(0, 0, 1), // Jan 1

@joebochill @huayunh What are the expected dates for each of these holidays?
Also, I may need to move the end date for new years to the 2nd to be inclusive of the 1st.

jeffvg:
> > export const NewYearsSchedule: Schedule = {
> > start: new Date(0, 11, 31), // Dec 31
> > end: new Date(0, 0, 1), // Jan 1
> 
> @joebochill @huayunh What are the expected dates for each of these holidays?
> Also, I may need to move the end date for new years to the 2nd to be inclusive of the 1st.

or maybe I don't understand the format here against the schedule.
when I see (0, 11, 31) I think Nov. 31

daileytj:
> > > export const NewYearsSchedule: Schedule = {
> > > start: new Date(0, 11, 31), // Dec 31
> > > end: new Date(0, 0, 1), // Jan 1
> > 
> > 
> > @joebochill @huayunh What are the expected dates for each of these holidays?
> > Also, I may need to move the end date for new years to the 2nd to be inclusive of the 1st.
> 
> or maybe I don't understand the format here against the schedule.
> when I see (0, 11, 31) I think Nov. 31

All good, the months start at 0 with the javascript date object so 0 is January.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth

huayunh:
You will need to do 1,0,1 for the New Years to wrap around 

And I suggest changing the date to Jan 2 (or even longer) for 
1. Jan 1 to be themed until 23:59
2. Making the theme change more high profile — Not many people will be checking Docit during New Years holiday, so if we extend it further more ppl will notice them)

huayunh:
And yeah, stupid JavaScript Date API design 🙄️ Don’t even get me started

huayunh (changes requested):
Overall, NEAT!

### Hanukkah
- I might flip the Hanukkah cursor upside down to match the shape of a regular mouse cursor

### Kwanzaa
- I would add text shadow to boost the text contrast
- I would change the secondary toolbar to red 900 on line35 to add a better transition between primary toolbar and secondary toolbar

### New Years 
- Nice! Had a big wooo when I saw the theme
- Adjust the date so that it wraps to the next year
- You need to copy & paste the dark theme background & text palette from the Halloween theme line 24 - line 32. Those provide better contrast. (Yes we will update our dark theme packages to match those colors)
- You also need to copy & paste the Halloween part in `index.css` to make the skeleton animation on the roadmap page and in-component code blocks work properly

For the background banner on the landing page:
- Have the cursor pointing top-left instead, and set its anchor point to `0 0`
- Change the title text to black would work better for the contrast
- I would also make the center of the banner slightly brighter to boost up the text contrast further

joebochill:
There a little overlap among these this year, so I suppose we'll want to run them:

- Hanukkah: Dec 9 through Dec 18
- Christmas: Dec 19 through Dec 25
- Kwanza: Dec 26 through Dec 30
- New Years: Dec 30 through Jan 2

And yeah, because of how the date logic works, "through" means we need to increment the end dates by one in the Date objects.

joebochill (commented):
**Hanukkah** 
- Looks absolutely dope. 
- I would drop the opacity some more on the tile image to make it a little more faded in the toolbar. 
- Also, the banner image is sick, but it's too big (1.5MB) — try cutting the dimensions to 960 in an image editor to try to get it down closer to 500kb. Or if there is something similar as an SVG that would cut the size too.

**Kwanza**
- I would try to avoid depicting _people_ in any of the graphics if we can.
- There are 7 'symbols of kwanza' — it would be cool if we could integrate some more of these into the tile...they would lend themselves very well to simple svgs I would think:
image
- Maybe use the flag for the cursor instead of an ear of corn?

**New Years**
- So sick
- Maybe just drop the opacity of the toolbar tile image some more so it doesn't interfere with the icons and text in the toolbar.
- Can we rotate the firework cursor so it points to the left like a default cursor (Huayun already mentioned this too)

Love what you did with making an assets folder for each theme...can you do the same for the existing ones and move the images into them (and make sure all the imports are updated)?

huayunh:
> I would try to avoid depicting people in any of the graphics if we can.

Cool with me as long as every human being is symbolic

src/app/assets/themes/kwanzaa/flag-cursor.svg
```diff
@@ -0,0 +1,15 @@
+<svg width="50" height="28" viewBox="0 0 50 28" fill="none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+<g clip-path="url(#clip0)">
+<rect width="50" height="28" fill="white"/>
+<rect width="50" height="28.125" fill="url(#pattern0)"/>
+</g>
+<defs>
+<pattern id="pattern0" patternContentUnits="objectBoundingBox" width="1" height="1">
+<use xlink:href="#image0" transform="scale(0.000520833 0.000925926)"/>
```
huayunh, line 8: Is this an image but pasted into an SVG file directly?

daileytj: Yeah, I resized it in figma and exported as an svg. Didn't think anything of it, is this a concern? I'm not totally familiar with all the filetype differences.

huayunh: It usually is a file size concern. SVG gives you vector graphs without pixelations; png is bitmap.

daileytj: PNG was significantly smaller in size so I swapped the cursor image out.

joebochill (approved):

jeffvg (approved):
all looks good to me

huayunh (approved):

PR #323, opened 2020-12-29
Feature/256 table of content

### LINKED ISSUE #256: TOC for long pages
## Describe the desired behavior

I want a table of content for loooong pages (#254 is currently 19648 px tall on desktops)

## Describe the current behavior

I tried doing 
```md
-   [Common Use Cases](#common-use-cases)
-   [Variations](#variations)
-   [Behaviors](#behaviors)
-   [Design Specifications](#design-specifications)
-   [Developers](#developers)
```

But it doesn't work. 

## Is this request related to a current issue?
No

## Additional Context
Deeplinking itself is still working, but it looks like when you have an internal link pointing to its own path, react router simply won't trigger anything.

huayunh:
Fixes #256

- Added a table of content to all the patterns page as PoC

## Screenshots

`md` and below:
image

`lg` (note the content is pushed to the side a little bit):
image

`xl`:
image

By the way, this is what I have in mind for the fixed positioned ToC:
image

I haven't implemented the scroll detector yet. Just want to collect some opinions first.

### 1
The logic was built inside "ScrollToTop", which uses `useEffect` on `pathname` and `hash`. So the page would scroll only when those two changes. So if you click on a toc link, then scroll around, and click on the same link, the page offset would stay the same.

### 2
I wanted each page component to automatically adjust itself based on the presence of a ToC. I tried to do a "useToC" hook, but couldn't think of a good enough hook logic. What I have in my mind so far is whenever a new page loads, an event "HAS_TOC, false" is dispatched. Then somewhere in ToC component's useEffect, another event is dispatched "HAS_TOC, true". Will these two conflict with each other? I haven't tried implementing it yet, just a regular React hook noob's panic.


joebochill (commented):
Functionally it seems pretty good. I'm not sure what number 2 means exactly...how do you want the page to update itself based on the presence of a TOC? I might be able to offer some thoughts if you can explain a bit more about the expected behavior.

I'm not sold on the design of this. The medium size one doesn't really offer much since it's off the top of the page as soon as you scroll a little bit or click on a link (maybe if we combine it with some sort of Back to Top fixed button). The large one feels out of place/far away from the content on a large monitor.

huayunh:
> how do you want the page to update itself based on the presence of a TOC?

Ah sorry, I didn't explain this carefully enough. This was meant for the `lg` screen size so that the page may nudge itself to the right and leave room for the ToC. 

Agree on the `xl` size ToC design — maybe we can have both ToC and content centered on the screen? How's this one?
image

joebochill:
> Ah sorry, I didn't explain this carefully enough. This was meant for the `lg` screen size so that the page may nudge itself to the right and leave room for the ToC.

You mean like what you're using the Spacer for now?
 
> Agree on the `xl` size ToC design — maybe we can have both ToC and content centered on the screen? How's this one?

I think that looks better, imo. I might even consider doing it on the right side of the content instead of left. Just a thought. Seems that's how a lot of sites do it.

huayunh:
> You mean like what you're using the Spacer for now?

Correct

joebochill:
> > You mean like what you're using the Spacer for now?
> 
> Correct

With the intent being that you don't want to have to specify `hasToc` for every page that has one?

You could probably do this with a hook + some sort of storage (redux or context). TOC would have a useEffect that runs once on first render to set a variable in the store (either boolean for hasTOC or we might need to store the name of the page to check against, not sure without trying it). And that hook would have a cleanup function that unsets the variable (this is the part I'm unsure of...there might be a race condition with one page TOC component unsetting the var while the next page TOC component is trying to set it...if you store the name of the page instead of a boolean, that should obviate that issue). Then the pageContent component reads from the store and check if the boolean is true (or if the stored page name matches the current page name) and then gives itself the spacer.

huayunh:
I can't think of a better design for down md, unless we try to tear down the drawer and do some magic to it

huayunh:
> this is the part I'm unsure of...there might be a race condition with one page TOC component unsetting the var while the next page TOC component is trying to set it...if you store the name of the page instead of a boolean, that should obviate that issue

Yeah this was the part I was unsure of, but storing the name is def a great idea here!

Should have made this a story instead of a bug :P

joebochill:
> Should have made this a story instead of a bug :P

🤣

huayunh:
Alright, I think I got the hooks part working.

One more trouble remaining for the scroll listener: the TOC apparently loads before all the other images, which suggests that the crawling of the `offsetTop` for all the anchors will happen after all the image loads.

This is no problem for the initial loading; we can conveniently do the `window.addEventListener('load'...)` thing. But since docit is a single-page app, the `window.onload` doesn't get fired when the pathname changes. So if I navigate to another page, the offsetTop crawling will happen first before all the images on the new page load. The TOC headline highlight thus gets inaccurate.

Solution # 1: make `initializeSectionOffsetTop` part of the scroll listener. Redo the section height stuff on every single scroll.

Solution # 2: Disable this scroll highlight feature.

Solution # 3: setInterval hack, re crawl the anchor point every 5 seconds

joebochill:
I messed around with it a little bit, and it seems like the least invasive way to get around it is with a settimeout:
image

jeffvg:
looks nice and good idea. Are we able to fix up the URL and highlighted page when we share a link and also when we select a navigation item and refresh the page to default back to base URL?

image

jeffvg:
not sure what this is... see this on every page

image

huayunh:
^ That is my fault from a previous PR. I am working on another branch to address that

huayunh:
Jeff, please see if this works a bit better; the console log thing has been fixed in #329

joebochill (approved):

src/app/hooks/useTOC.tsx
```diff
@@ -1,28 +1,14 @@
-import { useEffect, useState } from 'react';
+import { useEffect } from 'react';
 import { useDispatch } from 'react-redux';
-import { useLocation } from 'react-router-dom';
 import { TOGGLE_TOC } from '../redux/actions';
 
 export const useTOC = (toc: boolean): void => {
     const dispatch = useDispatch();
-    const { pathname } = useLocation();
-    const [prevTOCPathname, setPrevTOCPathname] = useState('');
```
joebochill: For future reference, we have a hook called 'usePrevious' that you can use to check a value against its previous value, if you run into a situation where you need something like that. I've used it in cases where I needed to compare a value in the redux store against a previous value (like going from loading = true to loading = false, or things like that).

jeffvg (approved):
looks good

PR #329, opened 2021-01-04
Global Search deleting characters will still search

### LINKED ISSUE #287: Global search should update whenever I hit Return Key
#### Describe the bug

Global search only update the first time I hit the return key. If I delete something and re-search, neither the URL nor the rest of the screen would update

Should always update when I hit return key

1. Go to global search
2. Type lskfsdklfjaslf
3. Observe that it searches for lskfsdklfjaslf
4. Delete a few chars, and hit return key again
5. Observe that nothing gets updated.

#### Environment

		
MacOS Chrome

huayunh:
Fixes #287


joebochill (approved):

jeffvg (approved):
looks good. bug is fixed and no errors

PR #340, opened 2021-01-14
Fix the query regex problem causing safari trouble

huayunh:
Fixes the thing we talked about in the chat

- Replace the look ahead syntax with some string slicing logic

To test, open up localhost in Safari.


daileytj (approved):
Looks good... Well done!

PR #345, opened 2021-01-21
Add new resources

emclaug2:
- Add angular auth workflow and all new templates to resources page.


src/__configuration__/resources/index.tsx
```diff
@@ -225,6 +232,67 @@ export const resources: ResourceBucket[] = [
...
+    {
+        name: 'CLI Templates',
+        description: 'Starter Templates for use with the PX Blue CLI',
+        applies: ['react', 'angular'],
+        items: [
+            {
+                name: '@pxblue/cra-template-authentication',
+                description: 'Javascript template for authentication workflow',
+                package: '@pxblue/cra-template-authentication-typescript',
+                repository: 'react-cli-templates',
```
joebochill, line 244: You may want to add the 'bugLabels' property to these items or else the bugs are going to be counted twice (when we start logging them) due to multiple templates being in a single repo. <--- This also means we will need to actually create issue labels in those repos — I was thinking something like:
- blank
- blank-ts
- routing
- routing-ts
- authentication
- authentication-ts

emclaug2: Good catch, I've updated docit to use these new labels and added the labels to the react and angular template repos.

joebochill (approved):

PR #346, opened 2021-01-26
Searchable FAQ

### LINKED ISSUE #330: Special Chars in Search are being replaced
#### Describe the bug
Special characters in the global search bar are being replaced by their URL encoded value. The issue compounds itself, because the special characters in the encoded string then get encoded again in the url.

Whatever the user types should not be modified (we can encode and do comparisons behind the scenes, but we should not be replacing the content that they type into the search field).

1. Go to doc it
2. Open the search
3. Search for "avatar="

#### Screenshots or links to minimum reproduction example
If I search for "avatar=" it gets replaced with this.
image

joebochill:
For PXBLUE-1802.
Fixes #330.

- Make the FAQ questions and answers searchable
    - Answers were already searchable (just needed to rebuild the index)
    - Added a find/replace line to pull out the questions before they get removed by the HTML tag remover line
- Simplify the implementation of the search results preview
- Fix a bug with encoding of special characters in the search string


src/app/pages/SearchFunction.ts
```diff
@@ -1,43 +1,53 @@
...
     }
-    const fullTextArray: string[] = fullText
+
+    fullText = fullText
         .replace(/import .*? from .*?;/gim, '')
         .replace(/\r\n/g, '\n')
         .replace(/\n/gim, ' ')
         .replace(/<!--.*?-->/g, ' ') // replace all comments (including keywords)
         .replace(/\[(.*?)\]\(.*?\)/g, '$1') // replace all the markdown links [text](url) into text
+        .replace(/<FAQExpander question={`(.*?)`}(.*?)>/gim, '$1') // pull out the FAQ question strings <FAQExpander question={''}>
```
huayunh, line 23: I can see how this might be extended to other uses, like searching for image captions, or searching for team member bios, or searching for design pattern developer info (if in the future we redesigned that sections to wrap in some components). Is it possible to do some kind of "comment tags" so that this can be applicable to other use cases?

```html
<!-- search: enabled --> blah <!-- search: disabled -->
```

joebochill: I agree that there are probably some more instances where we might like to include stuff in the search index. Not sure how we could effectively do it in a generic way though. The things that are currently getting stripped out are just props inside of HTML tags (we rip out the entire tag). I suppose we could do something like 'stringify all of the props, but I don't know if we really want all of them, probably just a few.

I'm not sure if something like:
```
<MyComponent /* search: enabled */ myProp={'some value'} /* search: disabled */ />
```
would even work. And if it did, would it be more convenient than just adding another condition to the indexer script for the specific prop we want?

huayunh:
```
<MyComponent /> some value
```

joebochill:
I got the caption searching to work with the following additional replace function:
```
.replace(/<ImageGrid.*?\/>/gim, (match) => {
            if(!match.includes('caption={')) return '';
            const str = match.replace(/<ImageGrid.*?caption={(.*?)}.*?\/>/gim, (match2, g1) => {
                let caption = g1.trim();
                if (caption.startsWith('[')) caption = caption.substr(1, caption.length - 2);
                return caption;
            })            
            return `(caption: ${str.trim()})`;
        })
```

But I'm not happy with the output in the search results. Having the image captions interrupting the flow of the main content is weird and confusing.

Maybe we should look into doing a bigger update for the search to have separate results for images/content instead.

src/app/pages/SearchFunction.ts
```diff
@@ -1,43 +1,53 @@
...
-    return [
-        keywordPosition > MAX_TEXT_LENGTH_HALF
-            ? fullTextArray
-                  .slice(keywordPosition - MAX_TEXT_LENGTH_HALF, keywordPosition + MAX_TEXT_LENGTH_HALF)
-                  .join(' ')
-            : fullTextArray.slice(0, MAX_TEXT_LENGTH).join(' '),
-        false,
-    ];
+    // otherwise return the initial text from the top of the page
+    console.log(fullText); // eslint-disable-line
```
huayunh, line 49: Clean up your code. I will review tomorrow.

src/app/pages/SearchFunction.ts
```diff
@@ -1,43 +1,53 @@
 import { Result } from '../../__types__';
 
 const MAX_RESULT = 10; // stop searching once we get 20 results
```
huayunh, line 3: Can you change this comment by the way 🤦‍♀️  my fault

huayunh (approved):
The rest of the code looks great!

PR #347, opened 2021-02-03
List dimension 56px -> 52px

huayunh:
Thy God of PX Blue; hallowed be your name. Bless me, for the sins I committed. The height of a dense list item was mistaken with the misleading of arrogance and ignorance. I firmly intend, for your help, to press the magic green button; and may your disciple be forgiven, the git history shall be written, and the shattered souls will lie in peace. 

## Penance
* Changed the dimension listed in the design spec from 56px to 52px, to match with the proper dimension of our Info List Item

## Other notes
* DSM has the correct size.
* Angular & React component libraries have the proper height


joebochill:
So our component libraries are currently using 52? What about the Figma stickersheet?

huayunh:
DSM now has the correct 52px size; its just a labeling error on docit.

Making it 52px also helps align with the height of a standard Material baseline table cell

joebochill (approved):

PR #353, opened 2021-02-10
Deploy the deep link fix for icons

joebochill:
- Fix the broken deep links for icons
- Move the initialization of the selection to after the icons list is loaded


huayunh:
The `isMaterial` is now a required parameter?

I thought it used to be: 
| |has a Material icon with matching name | has a PX Blue icon with matching name
|--|--|--|
|isMaterial=true | search | no search 
|isMaterial=false | no search | search
|no param | search for Mat icon by default, fall back to PX Blue icon | if no Mat icon, search for PX B icon, else don't toggle side bar

joebochill:
> The `isMaterial` is now a required parameter?
> 
> I thought it used to be:
> 
> has a Material icon with matching name	has a PX Blue icon with matching name
> isMaterial=true	search	no search
> isMaterial=false	no search	search
> no param	search for Mat icon by default, fall back to PX Blue icon	if no Mat icon, search for PX B icon, else don't toggle side bar

It's not required, but it is added to the URL whenever you click on an icon (for simplicity — we don't have to have an extra step to determine whether we should add it or not). It looks like the default is to find a PXB icon first (isMaterial = false).

Searching and selecting are separate right now. You can prepopulate the search using `?iconSearch=hello`, but we do not currently update the URL with the search term (I suppose we could, but is it a common use case to deep link someone to icon search results?).

emclaug2 (approved):

emclaug2:
When I was fixing the iconSearch bug I tinkered around to make it pre-populate the search with the iconSearch name,  I thought it was a little weird because the iconSearch name would sometimes filter to icons other than the one I was originally search for (due to metadata filters) so I left it out.

huayunh:
It is just for backward compatibility. We used to give out links of the form https://px-blue.github.io/style/iconography?icon=Accessibility. Just want those links to stay legit. Cool with me if we want to drop it.

joebochill:
> It is just for backward compatibility. We used to give out links of the form https://px-blue.github.io/style/iconography?icon=Accessibility. Just want those links to stay legit. Cool with me if we want to drop it.

Doing the fallback should be pretty trivial if we want to add it...we would just need to check if isMaterial===undefined and then check both the pxb set and the material set.

src/app/components/icons/IconBrowser.tsx
```diff
@@ -209,6 +203,13 @@ export const IconBrowser: React.FC = (): JSX.Element => {
     useEffect(() => {
         if (Object.keys(allIconsMap).length === 0) loadIcons();
         setIconsLoading(false);
+        // If loading from a query param, load the icon if it exists in the icon map.
+        if (iconQuery) {
+            if (allIconsMap[`${iconQuery}-${isMaterial ? 'material' : 'pxb'}`]) {
+                dispatch({ type: TOGGLE_SIDEBAR, payload: true });
+                setSelectedIcon(allIconsMap[`${iconQuery}-${isMaterial ? 'material' : 'pxb'}`]);
+            }
+        }
```
joebochill, line 212:
```suggestion
        if (iconQuery) {
            let iconKey = `${iconQuery}-${isMaterial ? 'material' : 'pxb'}`;
            if (allIconsMap[iconKey]) {
                dispatch({ type: TOGGLE_SIDEBAR, payload: true });
                setSelectedIcon(allIconsMap[iconKey]);
            }
            else if(isMaterial === undefined){
                iconKey = `${iconQuery}-material`;
                if (allIconsMap[iconKey]) {
                    dispatch({ type: TOGGLE_SIDEBAR, payload: true });
                    setSelectedIcon(allIconsMap[iconKey]);
                }
            }
        }
```

huayunh: I will make a separate branch for this.

huayunh (approved):

PR #359, opened 2021-02-17
Update to using React Component Library 5.0.0

huayunh:
- Update to using React component library 5.0.0
- I tried to do some InfoListItem props tricks to have nav items rendered as `Link`, but that causes some issues with parent items with a landing page (clicking on only the chevron icon will also cause the browser to open a new page)

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
         "@pxblue/colors": "^3.0.0",
         "@pxblue/colors-branding": "^3.0.0",
         "@pxblue/icons-mui": "^2.2.1",
-        "@pxblue/react-components": "^4.1.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
         "@pxblue/colors": "^3.0.0",
         "@pxblue/colors-branding": "^3.0.0",
         "@pxblue/icons-mui": "^2.2.1",
-        "@pxblue/react-components": "^4.1.1",
+        "@pxblue/react-components": "5.0.0",
```
joebochill, line 35:
```suggestion
        "@pxblue/react-components": "^5.0.0",
```

joebochill (approved):

PR #361, opened 2021-02-24
Make the item round again

huayunh:
I find the square activeBackgroundShape really awkward when nested items are involved — can be hard to see which one is getting selected. 

- useCallback to calculate the background color
- changed `activeItemBackgroundShape` to `round`

### Unrelated Stuff
I have another branch `feature/condensed-drawer` that plays around with the condensed drawer item variant, according to an earlier proposed drawer design. I know the design is not "final-final" yet, but as more and more pages are getting added to docit, I think it's time for docit to have a condensed drawer.
image

### DIFF
--- a/src/app/router/navigationDrawer.tsx
+++ b/src/app/router/navigationDrawer.tsx
@@ -60,6 +60,20 @@ export const NavigationDrawer = (): JSX.Element => {
         return convertedItems;
     }, []);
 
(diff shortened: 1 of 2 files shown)


joebochill (approved):
Not sure why you would want to switch to 'condensed' drawer style when there are more items since it actually takes up more space than the regular.

huayunh:
That has been redesigned to 48px now

PR #396, opened 2021-04-26
Add copy on click functionality

huayunh:
I found myself keep copying & pasting hex code and it annoys me that I cannot do that in one click.

- Add a copy-on-click functionality
- Shrink the hover area of those headlines, so that there would be no more unexpected misclicks.

### Copy color codes

This works for both hex and rgb.

| default | hover | clicked | your clipboard now is
| - | - | - | - |
| image | image | image |`#004b9e`
| image | image | image | `rgb(0, 75, 158)`

## Copy headlines
On desktop:
image

On a touch screen (defined by `@media (hover: none)`), the deeplink icon is always visible with a 0.25 opacity:
image

### DIFF
--- a/src/__configuration__/markdown/markdownMapping.tsx
+++ b/src/__configuration__/markdown/markdownMapping.tsx
@@ -94,7 +94,7 @@ const Headline: React.FC<Headline> = ({
                         copyTextToClipboard(`${window.location.origin}${window.location.pathname}#${hash}`);
                         setOnCopy(true);
                     }}
-                    style={{ position: 'relative', right: -8 }}
+                    style={{ marginLeft: 8 }}
                     size={'small'}
                     color={'primary'}
                 >
--- a/src/__configuration__/markdown/markdownMapping.tsx
+++ b/src/__configuration__/markdown/markdownMapping.tsx
@@ -9,6 +9,7 @@ import {
     Theme,
     createStyles,
     useTheme,
+    IconButton,
 } from '@material-ui/core';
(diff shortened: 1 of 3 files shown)


huayunh:
Don't know how to tame CircleCI. I'd wait until softwaretestinghelp fixes itself

src/__configuration__/markdown/markdownMapping.tsx
```diff
@@ -88,16 +85,21 @@ const Headline: React.FC<Headline> = ({
...
-                    titleAccess={'copy to clipboard'}
-                    fontSize={'inherit'}
-                    {...otherSvgIconProps}
-                />
+                <IconButton
+                    onClick={(): void => {
+                        copyTextToClipboard(`${window.location.origin}${window.location.pathname}#${hash}`);
+                        setOnCopy(true);
+                    }}
+                    style={{ position: 'relative', right: -8 }}
```
joebochill, line 97: What's with position relative? And why right: -8 instead of just a left margin?

joebochill (approved):

joebochill:
> Don't know how to tame CircleCI. I'd wait until softwaretestinghelp fixes itself

Yeah that's weird. It's throwing a 403 (Forbidden), but they load up just fine in the browser.

src/__configuration__/markdown/markdownMapping.tsx
huayunh: I had some issue with that, but it looks like it fixed itself after I dropped flexbox stuff. 

I changed it.

PR #401, opened 2021-05-03
Update prettier script and prettier everything

### LINKED ISSUE #395: Prettier Everything
Some of the files in the root should probably be formatted by prettier (e.g. package.json, scripts?).

#### Describe the current behavior
Only the src folder is prettiered.

#### Is this request related to a current issue?
No.

daileytj:
Fixes #395.

- Update prettier script and prettier everything
- Fix typo in cypress/support/readme

*NOTES:
- This hits every file now so we may want to update our @pxblue/prettier-config package to omit build folders so it skips those (not a big deal if we don't do this since build files generally aren't committed and the extra time to prettier those files is not significant). 
- Looks like prettier is removing escape characters in the link check .json files so we may want to ignore these files as well unless we just don't need those escape characters (if someone with more regex experience could take a look at this, that would be great).


daileytj:
The link check is saying there are dead links, but when I test the links manually they are working just fine ¯\_(ツ)_/¯. I thought this could be a result of those removed escape characters but this is happening on the community icons bug PR as well.

joebochill (approved):

PR #415, opened 2021-06-21
Update pattern contents for swipeable bottom sheet

joebochill:
For PXBLUE-2125

- Add links to recommended swipeable appbar libraries
- Update descriptions of the 3 types of bottom sheet (tried to keep the wording to a minimum since there are so many different things you could do with a swipeable bottom sheet).
- Add new graphic

#### Screenshots:
image


emclaug2 (approved):
Looks good to me.  Nice addition!

huayunh (changes requested):
- The map pin text should not be white.
- The black scrim in # 2 should cover the phone's status bar too. 

Where are "standard bottomsheets that do not block interactions" used in our applications?

joebochill:
> * The map pin text should not be white.

That would be an unexpanded stroke on the SVG when I shrunk it down. Will fix.

> * The black scrim in # 2 should cover the phone's status bar too.

Will fix.

> Where are "standard bottomsheets that do not block interactions" used in our applications?

I don't know that we have used them anywhere yet. Maybe the complex bottomsheet example in the monorepos? It was mentioned as an option before so I left it in there, but if it's not something we want to mention, I can take it out.

huayunh:
> Maybe the complex bottomsheet example in the monorepos? 

Ah, yes, that can be an option.... but only on desktop. We are more inclined to use a contextual action appbar on mobile.

Maybe a "file uploading" bottom sheet?

joebochill:
@huayunh something like this? I'm trying to keep the three examples in the same context to focus on the style of the bottomsheet over their contents.
image

jeffvg:
Do we need to mention anything about that all 3 implementations slide up from the bottom

huayunh:
While file uploading feels strange in a map context, this can work for me. You may also add a 6dp shadow around the outermost frame to make screens pop out better:

Screen Shot 2021-06-22 at 8 24 51 AM

joebochill:
> Do we need to mention anything about that all 3 implementations slide up from the bottom

image

huayunh:
image

jeffvg (approved):

huayunh:
don't rob my story :P 

Will review this afternoon

huayunh (commented):
# 1
#422.

# 2
Material Design calls it "expanding bottom sheet" instead of "Expandable". I do like yours better though. Do we want to make a small note on how MD calls it? I am fine either way.

# 3
Flow of text: I would introduce the three types of bottomsheet first, then talk about how the modal bottom sheet reshape responsively.

joebochill:
> Material Design calls it "expanding bottom sheet" instead of "Expandable". I do like yours better though. Do we want to make a small note on how MD calls it? I am fine either way.

What Material calls an expanding bottom sheet is much different from what we are describing (it's more like a FAB that opens a modal). Our definition is more like the default behavior of their standard bottom sheet: https://material.io/components/sheets-bottom#behavior

huayunh (approved):

PR #417, opened 2021-06-21
Scaffold for R20 release

huayunh:
Some very basic bullet points, waiting to be changed when this sprint is over.

### DIFF
--- a/src/__configuration__/roadmap/index.ts
+++ b/src/__configuration__/roadmap/index.ts
@@ -1,6 +1,6 @@
 import { Quarter, Release } from '../../__types__';
 
-export const CURRENT_RELEASE: Release = 'R20';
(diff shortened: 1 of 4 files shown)


src/__configuration__/roadmap/index.ts
```diff
@@ -1,6 +1,6 @@
 import { Quarter, Release } from '../../__types__';
 
-export const CURRENT_RELEASE: Release = 'R20';
+export const CURRENT_RELEASE: Release = 'R21';
```
joebochill, line 3: We usually leave this as R20 for a week or so after the release so people will see the details of what we did (and it also gives us a chance to populate R21). I'll switch it back when I go in and update the notes.

joebochill (approved):

PR #425, opened 2021-06-24
Move indexer to yarn indexer script

### LINKED ISSUE #424: yarn indexer on build
Do `yarn indexer` right before the build

huayunh:
Fixes #424 

- Move yarn indexer out of precommit and move into yarn build
- This means that we always have an outdated local global search (unless someone triggers it manually). Do we want this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "private": true,
     "scripts": {
         "start": "node scripts/start.js",
-        "build": "node scripts/build.js",
+        "build": "yarn indexer && node scripts/build.js",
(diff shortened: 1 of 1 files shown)


joebochill:
I'm okay with this change. We can always switch it back if we need to. But this will guarantee that the deployed site is always up to date. When doing stuff with search locally, you will just need to remember to manually run the indexer when necessary. This should also keep future PRs a bit easier to read without always having to see the index files changed.

joebochill (approved):

PR #427, opened 2021-06-28
Feature/download icons

ektasawant:
Fixes PXBLUE-1680 .

- Added SVG and PNG download button
- Change SVG color size as per selection
- 

#### To Test:
-


src/app/components/icons/IconDrawer.tsx
```diff
@@ -150,16 +172,82 @@ export const IconDrawer: React.FC = () => {
...
+                                    ) : (
+                                        <Select
+                                            labelId="icon-size-select-label"
+                                            id="icon-size-select"
+                                            value={iconSize}
+                                            onChange={handleChange}
+                                        >
+                                            <MenuItem value={24}>24dp</MenuItem>
+                                            <MenuItem value={48}>48dp</MenuItem>
+                                        </Select>
```
joebochill, line 201: instead of conditionally rendering the entire Select, you can conditionally render the two extra menu items:

```
<Select
    labelId="icon-size-select-label"
    id="icon-size-select"
    value={iconSize}
    onChange={handleChange}
>
    {selectedIcon.isMaterial && <MenuItem value={18}>18dp</MenuItem>}
    <MenuItem value={24}>24dp</MenuItem>
    {selectedIcon.isMaterial && <MenuItem value={36}>36dp</MenuItem>}
    <MenuItem value={48}>48dp</MenuItem>
</Select>
```

joebochill: Same for the color picker

src/app/components/icons/IconDrawer.tsx
```diff
@@ -150,16 +172,82 @@ export const IconDrawer: React.FC = () => {
...
                                 </Button>
+                                <Button
+                                    variant="contained"
+                                    color="primary"
+                                    onClick={(): void => downloadPng(selectedIcon, iconColor, iconSize)}
+                                    startIcon={<GetApp />}
+                                >
+                                    PNG
+                                </Button>
                                 {!selectedIcon.isMaterial && (
```
joebochill, line 251: You can remove this text block about opening in a new window now as it is no longer applicable

src/app/components/icons/IconDrawer.tsx
```diff
@@ -150,16 +172,82 @@ export const IconDrawer: React.FC = () => {
...
+                                        <Select
+                                            labelId="icon-color-select-label"
+                                            id="icon-color-select"
+                                            value={iconColor}
+                                            onChange={handleColorChange}
+                                        >
+                                            <MenuItem value={'Black'}>Black</MenuItem>
+                                            <MenuItem value={'White'}>White</MenuItem>
+                                            <MenuItem value={'Blue'}>Blue</MenuItem>
+                                            <MenuItem value={'Gray'}>Gray</MenuItem>
```
joebochill, line 226: You could make the values here the explicit color that we want to use (so you won't need any logic to look them up in the download function):

```
<MenuItem value={Colors.black[500]}>Black</MenuItem>
```

src/app/components/icons/IconDrawer.tsx
```diff
@@ -69,6 +80,8 @@ export const IconDrawer: React.FC = () => {
     const history = useHistory();
     const dispatch = useDispatch();
     const classes = useStyles(theme);
+    const [iconSize, setIconSize] = React.useState(24);
+    const [iconColor, setIconColor] = React.useState('Black');
```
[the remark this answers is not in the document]
joebochill: Or we could leave color as a string if we use the literal color values (see comment below).

src/app/components/icons/utilityFunctions.tsx
```diff
@@ -15,21 +17,58 @@ export const getMuiIconName = (filename: string): string => {
     return muiName;
 };
 
+export const changeSvgColorAndSize = (svg: string, color: string | undefined, size: number | undefined): string => {
+    let newSvg = svg;
+    
+    if (color) {
+        // @ts-ignore
+        const colorCode = color === 'White' ? Colors[color.toLowerCase()][50] : Colors[color.toLowerCase()][500];
```
joebochill, line 25: You can skip this color lookup logic if we use the literal color values in the Select (see comment above). Or you could pull this out to another function that will be easier to extend in the future if we want to do more colors:

```
const getColorCode = (color:string) => {
    const lowerColor = color.toLowerCase();
    switch(lowerColor){
        case 'black':
        case: 'gray':
        case: 'blue':
            return Colors[lowerColor as (keyof typeof Colors)][500];
        case 'white':
            return Colors.white[50];
        default:
            return Colors.black[500];
    }
}
```

src/app/components/icons/utilityFunctions.tsx
```diff
@@ -15,21 +17,58 @@ export const getMuiIconName = (filename: string): string => {
...
-export const downloadPng = (icon: IconType): void => {
-    window.open(`//fonts.gstatic.com/s/i/materialicons/${icon.iconFontKey}/v6/black-24dp.zip?download=true`, '_blank');
+export const downloadPng = (icon: IconType, color: string, size: number): void => {
+    if (icon.isMaterial) {
+        window.open(`//fonts.gstatic.com/s/i/materialicons/${icon.iconFontKey}/v6/${color.toLowerCase()}-${size.toString()}dp.zip?download=true`, '_blank');
+    } else {
+        const colorName = color === 'White' ? `${color}50` : `${color}500`
+        window.open(`https://raw.githubusercontent.com/pxblue/icons/dev/png/png${size}/${getSnakeCase(icon.name)}_${colorName.toLocaleLowerCase()}_${size}dp.png?download=true`, '_blank');
+    }
 };
```
joebochill, line 72: Can we download PNGs the same way we are doing the SVGs to avoid opening them in a new window? And just skip the color/size function.

src/app/components/icons/IconDrawer.tsx
```diff
@@ -69,6 +80,8 @@ export const IconDrawer: React.FC = () => {
     const history = useHistory();
     const dispatch = useDispatch();
     const classes = useStyles(theme);
+    const [iconSize, setIconSize] = React.useState(24);
```
joebochill, line 83: We will need to be careful here if somebody is switching between Material Icons and PX Blue icons with a size or color selected that is not supported by the new icon. We may need to reset this selection each time the icon is changed (if the icon class is different).

src/app/components/icons/IconDrawer.tsx
joebochill: I just realized this would overcomplicate the logic in the PNGs, so we can probably stick with what we have (maybe just lowercase the `values` so we don't have to do it later).

src/app/api/index.tsx
```diff
@@ -98,3 +106,31 @@ export const getRoadmap = async (release: Release): Promise<RoadmapBucket[] | un
...
+        if (response && response.status === 200) return response.data;
+        return undefined;
+    } catch (thrown) {
+        if (axios.isCancel(thrown)) {
+            // request canceled
+            return undefined;
+        }
+        return undefined;
+    }
+};
```
joebochill, line 136: You could combine these if you want:

```
export const getSvg = async (name: string, family: 'material" | 'pxblue'): Promise<string | undefined> => {
    ...
    const response = family === 'pxblue' ? await pxblueIcons.get(...) : icons.get(...);
    ...
}
```

src/app/components/icons/IconDrawer.tsx
```diff
@@ -69,6 +80,11 @@ export const IconDrawer: React.FC = () => {
     const history = useHistory();
     const dispatch = useDispatch();
     const classes = useStyles(theme);
+    type IconSize = 18 | 24 | 36 | 48;
+    type IconColor = 'Black' | 'Blue' | 'Gray' | 'White';
```
joebochill, line 84: I would suggest defining these in the shared types fie (https://github.com/pxblue/doc-it/blob/dev/src/__types__/index.ts) and export them so we can use them in the utility functions.

src/app/components/icons/IconDrawer.tsx
```diff
@@ -77,6 +93,11 @@ export const IconDrawer: React.FC = () => {
         dispatch({ type: TOGGLE_SIDEBAR, payload: false });
     };
 
+    useEffect(() => {
+        setIconSize(24 as IconSize);
+        setIconColor('Black' as IconColor);
```
joebochill, line 98: only change these if we go from a material icon to a pxblue icon. You can check this with the usePrevious hook (should already be available in doc-it, I think):

```
const previousSelectedIcon = usePrevious(selectedIcon);
...
useEffect(() => {
    if(previousSelectedIcon && previousSelectedIcon.isMaterial !== selectedIcon.isMaterial){
        { /* reset the size */ }
    }
}, [selectedIcon);
```

src/app/components/icons/utilityFunctions.tsx
```diff
@@ -15,21 +17,85 @@ export const getMuiIconName = (filename: string): string => {
     return muiName;
 };
 
-// Can be Material or PX Blue icons
-export const downloadSvg = (icon: IconType): void => {
-    if (icon.isMaterial) {
-        window.open(
-            `https://fonts.gstatic.com/s/i/materialicons/${icon.iconFontKey}/v6/24px.svg?download=true`,
-            '_blank'
+const getColorCode = (color:string): string => {
```
joebochill, line 20:
```suggestion
const getColorCode = (color: IconColor): string => {
```

src/app/components/icons/utilityFunctions.tsx
```diff
@@ -15,21 +17,85 @@ export const getMuiIconName = (filename: string): string => {
...
+        case 'blue':
+            return Colors.blue[500];
+        case 'white':
+            return Colors.white[50];
+        default:
+            return Colors.black[500];
+    }
+}
+
+export const changeSvgColorAndSize = (svg: string, color: string | undefined, size: number | undefined): string => {
```
joebochill, line 36:
```suggestion
export const changeSvgColorAndSize = (svg: string, color: IconColor | undefined, size: IconSize | undefined): string => {
```

src/app/components/icons/utilityFunctions.tsx
```diff
@@ -15,21 +17,85 @@ export const getMuiIconName = (filename: string): string => {
...
+    const image = await fetch(imageSrc) || ''
+    const imageBlog = await image.blob() || ''
+    const imageURL = URL.createObjectURL(imageBlog)
+  
+    const link = document.createElement('a');
+    link.href = imageURL;
+    link.download = imageName;
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
```
joebochill, line 77: Can we functionalize this element creation since the logic is the same for SVGs and PNGs

src/app/components/icons/utilityFunctions.tsx
```diff
@@ -15,21 +17,85 @@ export const getMuiIconName = (filename: string): string => {
...
+}
+// Material or PX Blue SVG icons
+export const downloadSvg = async (icon: IconType, color: string, size: number): Promise<void> => {
```
joebochill, line 80:
```suggestion
export const downloadSvg = async (icon: IconType, color: IconColor, size: IconSize): Promise<void> => {
```

src/app/components/icons/utilityFunctions.tsx
```diff
@@ -15,21 +17,85 @@ export const getMuiIconName = (filename: string): string => {
...
+        // create a placeholder dom element and attach the new material icon svg download handler to it
+        createDownloadSvgElement(icon, iconData, color, size);
+    } else {
+        const iconData = await getPxblueSvg(getSnakeCase(icon.name).toLowerCase()) || '';
+        // create a placeholder dom element and attach the new pxblue icon svg download handler to it
+        createDownloadSvgElement(icon, iconData, color, size);
+    }
+};
+// Material or PX Blue PNG icons
+export const downloadPng = (icon: IconType, color: string, size: number): void => {
```
joebochill, line 92:
```suggestion
export const downloadPng = (icon: IconType, color: IconColor, size: IconSize): void => {
```

joebochill:
Looking really good, just a few cleanup items. Make sure you run prettier and lint.

joebochill (approved):

huayunh:
good to have: #429

PR #445, opened 2021-07-21
Talk about safe area in patterns.

huayunh:
- Add a few words and pics about safe area.

#### Screenshots:

image


joebochill (approved):
Interesting choice of branch name: 
image

Content looks good..I'm not 100% on the placement (these guidelines apply to the regular app bar and the collapsible app bar too). But it's probably fine...putting it at the very end feels weird too. I like it in the behaviors section for Navigation, but we don't have that for app bar.

huayunh:
> putting it at the very end feels weird too

Yeah I struggled too

PR #446, opened 2021-07-22
Add info about animations

huayunh:
- Add a little blurb about animations

#### To Test:
- New text located here: http://localhost:3000/style/illustrations


src/docs/style/illustrations.mdx
```diff
@@ -68,6 +69,12 @@ To maximize accessibility, use online tools (such as [Coblis](https://www.color-
     regularWidth
 />
 
+### Animation
+
+<ImageGrid images={[AnimationImage]} caption={'An animated empty state illustration.'} />
+
+Animations can draw attention to important parts. When combined with instruction text, they can provide a very interactive onboarding experience. We recommend [Lottie](https://lottiefiles.com/) to export and implement your animation.
```
joebochill, line 76: "important parts" of what? The illustration, or the UI?
```suggestion
Animations can draw attention to important parts of the user interface. When combined with instructional text, they can provide a very interactive onboarding experience. We recommend using [Lottie](https://lottiefiles.com/) to implement and export your animation.
```

What's good about Lottie versus something like Figma which is already a standard tool?

huayunh: - Lottie is highly recommended by Bryan Bunns, who suggests that this is an industry-standard.
- You can do the SVG part in Figma (or Adobe Illustrator, or other SVG tools), but then to animate the SVG and implement it inside the project, you would want to try out Lottie, which gives you an actual SVG animation rather than a GIF.
- I'd want to explore further and see what Lottie can do for a flex-time story

huayunh: > "important parts" of what? The illustration, or the UI?

In that particular example, the animation is meant to remind users to look for devices with blinking lights on

joebochill: Good to know about Lottie...maybe we soften the language a bit until we've had a chance to play with it and say something like "You can use tools such as Lottie...". And then after we are a little more familiar we can upgrade that to a recommendation.

For the important parts, I would just add a small clarifying blurb. Like "draw attention to important parts of the graphic that are not easily conveyed with a static image"

joebochill (approved):

PR #448, opened 2021-07-23
Add info about drawer header turning blank space

### LINKED ISSUE #437: Update navigation section so that ppl understand better how to

Update Docit navigation pattern to include information about conditionally showing the drawer NavGroup title to aviod the awkward white space thing when the drawer collapsed.

Add new content to doc-it describing the behavior of drawer nav group title when the drawer is collapsed. (hide the space vs. keep the space)

#### Additional Context
This is something that was brought up and discussed with the Green Motion team.

huayunh:
Fixes #437 

- Added a gif to warn ppl about using persistent drawer with drawer nav header, and a pic to instruct ppl to get around it

#### To Test:
- http://localhost:3000/patterns/navigation#persistent


src/docs/patterns/navigation.mdx
```diff
@@ -131,6 +133,13 @@ A persistent drawer is an always-visible drawer that can be opened and closed (e
         }
         regularWidth
     />
+    <ImageGrid
+        images={[PersistentDrawerPitfallImg, CombinedDrawerImage]}
+        caption={[
+            'Be careful when using persistent drawer with drawer item group headers (where it says "GROUP 1" and "GROUP 2" in the image), as they turn into blank spaces when the drawer collapses.',
+            `One way to get around this issue is to use different drawer variants at different view ports. For example, you can use a permanent drawer on desktop screens (#1) and switch to a persistent drawer without header text or a navigation rail on tablets (#2).`,
+        ]}
```
joebochill, line 141:
```suggestion
            'When using a persistent drawer with NavGroup titles (where it says "GROUP 1" and "GROUP 2" in the image), the titles become blank spaces when the drawer collapses by default.',
            `If you prefer a look without these spaces, you can conditionally hide the title when the drawer collapses or use different drawer variants at different view ports. For example, you can use a permanent drawer on desktop screens (#1) and switch to a persistent drawer without header text or a navigation rail on tablets (#2).`,
        ]}
```

joebochill: If it's an "issue" (versus a stylistic choice), we should fix it rather than describing a workaround

huayunh: It is more of a design issue rather than a technical one...

joebochill: I would avoid calling it an 'issue' then and say something along the lines of what I suggest above

huayunh:
Didn't sign out , oof

joebochill (approved):

PR #455, opened 2021-08-17
Rm joe; fix broken MD links

huayunh:
Fixes PXBLUE-2420

## Changes proposed in this Pull Request
- Remove Joe from current maintainers
- Fix broken markdown links

## Screenshots

image


src/__configuration__/contributors/contributors.ts
```diff
@@ -106,4 +107,15 @@ export const contributors: Contributor[] = [
         description: 'Icons',
         image: RenatoMartins,
     },
+
+    {
+        name: 'Joseph Boyle',
+        description: `Front-End Developer & Strategy`,
+        image: JosephBoyle,
+        info: `Forsaking all others, Joe has been with PX Blue since its inception. A double-threat with a background in computer science and user experience, he's not afraid to drop 100+ comments on your most recent PR. Befriend at your own risk.`,
```
emclaug2, line 115: How are the fields `info` & `contacts` used here?

emclaug2:
Looks good just see one comment.

src/__configuration__/contributors/contributors.ts
huayunh: Can go either way; they share the same type, but info and contacts don't get rendered

emclaug2: I'm thinking these should be two different types if those other fields aren't used.

emclaug2 (approved):

Issue #456, opened 2021-08-19
In dev mode, auto refresh cause IconDrawer accessed before initialization


huayunh:
#### Describe the bug
```
ReferenceError: Cannot access 'IconDrawer' before initialization
```

#### What is the expected behavior?

Auto-refresh shall work

#### What are the steps to reproduce?
1. `yarn start`
2. go to the icon-library (formerly iconography) /style/icon-library page, observe that everything gets rendered properly
3. randomly change something
4. waiting for the server to auto-refresh
5. observe RSOD
6. refresh using browser refresh
7. everything back to normal

#### Screenshots or links to minimum reproduction example

![image](https://user-images.githubusercontent.com/8997218/129994113-61f6c687-b62b-41ea-8863-1eb0ec3c7efb.png)

#### Environment

Mac Big Sur Chrome
		
#### Anything else to add?

This is only in dev. Everything in build works fine.

I can live with refreshing... just annoying.

huayunh:
Leave this open

bkarambe:
@huayunh I am not able to reproduce this issue. Can you please confirm, whether it is already been fixed?

huayunh:
The issue still exists in dev, but it looks like it has been miraculously fixed in the MUIv5 branch.

PR #457, opened 2021-08-19
Modify RN icon instructions

huayunh:
Fixes PXBLUE-2370

## Changes proposed in this Pull Request
- Added RN instructions
- Tweaked divider styles
- Identified #456

## Screenshots
A Material icon:
image

A PXB icon:
image

## To Test
- http://localhost:3000/style/iconography?icon=EatonTwoTone&isMaterial=false


joebochill:
I would not show the Icon Family section here. That is one specific way of icon usage within the component library and we have a whole document that goes into detail about all of the different ways to do that. The way it sits now it looks like you are rendering an object to the screen which won't work. (also the version of the component library supporting this format has not been released to the public yet).

jeffvg (approved):
looks good

joebochill (approved):

PR #471, opened 2021-10-01
Update links to account menu, appbar, forms and loading

huayunh:
- Updated React's link to the three design patterns we have implemented: account menu, appbar and forms
- Disabled Angular due to technical troubles with StackBlitz
- There might still be a few React links in other design patterns pointing to the wrong place, need to double check before we can release this into R21
- Feel free to commit to this branch directly


src/__configuration__/resources/index.tsx
```diff
@@ -324,7 +324,7 @@ export const resources: ResourceBucket[] = [
                 description: 'Web design patterns made using the Angular framework',
                 repository: 'angular-design-patterns',
                 applies: ['angular'],
-                demoUrl: 'https://stackblitz.com/github/pxblue/angular-design-patterns',
```
huayunh, line 327: this is a great catch, thanks Ekta

emclaug2 (approved):

PR #474, opened 2021-10-08
Out folder already taken

emclaug2:
- Fixes the broken dev build in CI.
- The `out` folder name was already taken we deployed to the px-blue dev doc-it, so use a diff folder name.


daileytj (approved):

huayunh:
I guess the best way to test is to merge this :)

PR #483, opened 2021-12-02
Use session storage to hide banner on refresh

huayunh:
- use `window.sessionStorage` to cure my OCD

#### To Test:
- yarn start
- Observe that the banner gets rendered
- Dismiss it
- Refresh
- Observe that banner is gone. In the current dev branch, the banner would get restored upon refresh.
- Open localhost in another tab
- Observe that the banner comes back


andrupu (approved):
Works like magic.

PR #491, opened 2021-12-17
Release Notes go PX Blue

huayunh:
- Make release notes say PX Blue for R21 and before


daileytj (approved):
Approved, assuming we're cool with all the broken links.

PR #502, opened 2022-01-06
Show Do's and Dont's in a better way

huayunh:
- Explicitly call out do and don't and avoid and caution
  - There is one image in the app bar pattern that I didn't do... can't think of a good way to treat it
  - 

## Screenshots:

Before: 
image

After:
image


jeffvg:
> A link (https://www.guru99.com/find-element-selenium.html) in development/testing.mdx is broken. @jeffvg any suggestions for what to replace here?

the link seems ok https://www.guru99.com/find-element-selenium.html but has a "TON" of adverts on it now....
let me look around to see if a better one is out there
we has issues last night with the medium.com links as well.

jeffvg:
@huayunh lets use this link, it seems to have the same info with less advertising 

https://www.browserstack.com/guide/findelement-in-selenium

daileytj (approved):
Looks good to me!

PR #504, opened 2022-01-06
Ensure EV Plug instructions and links all works

### LINKED ISSUE #486: Iconography page giving wrong instructions for icon fonts when there are consecutive numbers
### Describe the bug

For icons named `ev_plug_iec_60309_2_xxx` (there are four of them), docit is omitting the dash in `_2` part of the name, and the icon name is listed to be `ev_plug_iec_603092_xxx`

### What is the expected behavior?

It should name correctly. This applies to icon fonts and SVG instructions.

### What are the steps to reproduce?
1. docit needs to upgrade to icons-mui v2.8.0^ first
2. Checkout an icon named as described (for example, /style/iconography?icon=EvPlugIec603092Three32&isMaterial=false)
3. Check the instructions in the side panel

### Screenshots or links to minimum reproduction example
Highlighted instructions should be identical
image

huayunh:
Fixes #486

- Fixed EV Plug code snippet
- Ensure that EV Plug icons SVG and PNGs can both download properly
- Fixed a few places that still has traces of pxblue

#### Screenshots:

image

#### To Test:
- Verify that the four EV plugs mentioned in the bug report now have correct usage instructions
- Verify that the copy button for those code snippets copy correctly
- Verify that the Icon Drawer display the icon title correctly
- Verify that everything downloads and opens correctly. (Icons with wrong names can still be downloaded, but the file is corrupted and won't open)
- Verify that other regular icons (both blui icons and Material icons) are left untouched

### DIFF
--- a/src/app/components/icons/IconBrowser.tsx
+++ b/src/app/components/icons/IconBrowser.tsx
@@ -150,7 +150,7 @@ const loadIcons = (): void => {
 
         const icon: IconType = {
             name: iconKey,
-            iconFontKey: iconDetails.name,
+            iconFontKey: iconDetails.filename.slice(0, -4),
             type: 'Filled',
             isMaterial: false,
             tags: iconDetails.tags || [],
--- a/src/app/components/icons/IconDrawer.tsx
+++ b/src/app/components/icons/IconDrawer.tsx
@@ -24,7 +24,7 @@ import { EmptyState, Spacer } from '@brightlayer-ui/react-components';
 import { GetApp, Close } from '@material-ui/icons';
 import { Pxblue } from '@brightlayer-ui/icons-mui';
 
-import { unCamelCase } from '../../shared';
+import { snakeToTitleCase } from '../../shared';
 import { emptyIcon } from '.';
 import { downloadPng, downloadSvg } from './utilityFunctions';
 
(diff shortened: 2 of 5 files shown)


src/app/components/icons/IconBrowser.tsx
```diff
@@ -150,7 +150,7 @@ const loadIcons = (): void => {
 
         const icon: IconType = {
             name: iconKey,
-            iconFontKey: iconDetails.name,
```
emclaug2, line 153: Was there a reason for changing the key used for the BLUI icons?  Fine with the change regardless but wondering if iconDetails.name isn't guaranteed as unique.

emclaug2 (approved):
Changes worked locally. Changes look reasonable, less functions calls in the jsx looking good.

src/app/components/icons/IconBrowser.tsx
huayunh: I think the intent here is, icon.name is the "MUI name", while iconFontKey would be the filename without the ".svg" part. So if I have an icon called "px-blue-icon.svg", the `icon.name` field would be `PxBlueIcon`, and the `icon.iconFontKey` should store `px-blue-icon`.

`iconDetails.name` stores the filename in a human-readable way, in Title Case. So using the example above,  `iconDetails.name` would be `PX Blue Icon`. However, this thing has some legacy issues with it and can cause some confusion, so it is not actively used anywhere.

PR #512, opened 2022-01-17
Remove "X" from the search field

huayunh:
- Remove the browser default "X" from the search field
- Just a small thing but it bugs me a little...
- 

#### Screenshots:

Before: 

https://user-images.githubusercontent.com/8997218/149817595-4a767e15-435e-460e-9740-7273a22166b2.mov

After:

https://user-images.githubusercontent.com/8997218/149817606-a3455546-f466-417b-87ae-28462bd9286b.mov

#### To Test:
- search for something
- hover your mouse in the search field
- ensure that the x does not appear

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -34,6 +34,13 @@ q:after {
     padding: 0;
 }
 
+input[type='search']::-webkit-search-decoration,
+input[type='search']::-webkit-search-cancel-button,
+input[type='search']::-webkit-search-results-button,
+input[type='search']::-webkit-search-results-decoration {
+    display: none;
+}
+
 blockquote > p:last-child {
     margin-bottom: 0;
 }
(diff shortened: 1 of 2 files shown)


jeffvg:
I think this is a chrome thing and I don't see it in FireFox. Right now in this branch I can search without seeing the double close now but the search stops and places a "?" in the url without the entire route search "?&search=theme" weird one


https://user-images.githubusercontent.com/37150565/149823757-e52059e6-d7e9-43d0-91a4-e21a23735f90.mov

src/index.css
```diff
@@ -34,6 +34,13 @@ q:after {
     padding: 0;
 }
 
+input[type='search']::-webkit-search-decoration,
```
emclaug2, line 37: Looks good to me.

emclaug2:
> I think this is a chrome thing and I don't see it in FireFox. Right now in this branch I can search without seeing the double close now but the search stops and places a "?" in the url without the entire route search "?&search=theme" weird one

This is specific to firefox?  If the search is being impacted in some way could you log an issue for this?  

Approving this because the X is gone for me on Chrome.

emclaug2 (approved):

huayunh:
Great catch, jeff! Removed the "form" HTML tag that is clearly not compatible with single page apps

PR #523, opened 2022-02-17
Publish Docit

emclaug2:
- New Typography naming convention
- Aprils Fools Day Theme
- Hide Eaton logo on Icons Page
- Colors Accessibility Improvements


src/app/components/typography/TypographyTable.tsx
```diff
@@ -109,14 +109,25 @@ export const TypographyTable = (): JSX.Element => {
...
+        if (variant.startsWith('h')) {
+            const num = variant.charAt(1);
+            if (num === '1') {
+                return `H1`;
+            }
+            return `Headline${num}`;
+        }
+        return variant.charAt(0).toUpperCase() + variant.slice(1);
+    }, []);
+
```
joebochill, line 122: Why is H1 not Headline 1 like in Material?

Can we put a space between "Headline" and the number?

image

emclaug2: H1 is not "Headline 1" because the font-size is so big (and name so long) that it causes an overflow in the table which doesn't look so great.  I'll make a small PR to add the space.

emclaug2: I abused my admin privs and pushed directly to dev.

huayunh: An overflowing table hides a column which should not be hidden…

And the phrase “headline” also, theoretically speaking, have some alternative meanings in Angular Material.

But, if you insist changing everything to “Headline #”, I suggest taking out the column saying “open sans” as it does not really contribute much information (although I still love seeing all information together)

emclaug2: It's not really a wrap, it's an overflow like @huayunh mentioned; maybe we can talk about this in office hours and figure out a good solution.  I'll wait til we have consensus before making any code changes.

joebochill: If I change H1 to Headline 1, i see a wrap, not an overflow (no hidden columns):
image

And you'll run into overflow on smaller screens even if you call it H1...that's unavoidable with tables.

Headline 1 is (~90%) less confusing than H1/H2/H3/H4 because there is no `<headline>` element that you can mistakenly use. Thinking that using an `<h1>` will get you a BLUI Headline 1 style in Angular Material is what sparked this issue to begin with.

What about something like:
image


I'm also okay with dropping the typeface column as a quick workaround for now.

huayunh: ^ Can do this for a smaller view port

huayunh:
We can remove the space in "Headline 1" to prevent the text wrap

We can also leave it there, enforce a white space no-wrap rule, and drop down the horizontal table cell padding to 8px.

emclaug2:
> We can also leave it there, enforce a white space no-wrap rule, and drop down the horizontal table cell padding to 8px.

I opted to try this, but ran into some issues just trying to apply custom style overrides to the Table.  I'll bring this up tomorrow in office hours so we can get this wrapped up asap.

emclaug2:
This PR should be good to go now, but we need one more approval on it @jeffvg @daileytj @joebochill .

daileytj (approved):
Looks good... April Fools day theme is pretty solid too

huayunh:
> April Fools day theme is pretty solid too

Lol, I thought this could sneak in without you noticing it so that you can get tricked too

PR #533, opened 2022-03-16
Goodbye Ionic

emclaug2:
- Removes all references to Ionic


jeffvg:
is there a plan to remove the "hidden" ionic entries in navigation and index later?

emclaug2:
> is there a plan to remove the "hidden" ionic entries in navigation and index later?

I was thinking to keep them as a reference; we've hid pages before so I treated this the same way.

jeffvg (approved):
looks good to me

PR #547, opened 2022-04-14
Update announcement bar using blui db

ektasawant:
Fixes 3041 .

- Update announcement bar details using blui-database
- show / hide as per start and end date
- move announcementBar code from router.

#### Screenshots:
Screenshot 2022-04-14 at 4 38 48 PM

#### To Test:
- Checkout to `feature/3041-update-announcement-data` in `blui-database`. I am using json-server to host json 
- run json-server --watch announcement.json  --port 3001 in `deployed/doc-it`.
- In doc-it run yarn start.


emclaug2:
@ektasawant Could you elaborate more about the schema you're proposing in `Announcement.json`?

Here's a couple things I've noticed so far:
1. When the datasource is inaccessible, the banner still displays but with empty content.
2. The link "how to migrate" in the demo appears as a link, but it dismisses the banner.  I'm confused about how the link is populated since there's no URL present in the `Announcement.json` file.
3. How would you handle the situation where we need multiple links in the banner?  e.g. `We've related updated to [R24](link1)! Checkout our release presentation on [Youtube](link2).`
4. Do you have any thoughts around what happens when the banner has previously been dismissed in local storage, but the `Announcement.json` content has changed.  I'm thinking the localstorage would currently disregard the incoming data & hide the banner.

jeffvg:
Just noticing this when scrolling.

https://user-images.githubusercontent.com/37150565/164252492-bbedfa95-55aa-42d5-bbbc-8ee1c17639c1.mov

jeffvg:
also wondering how we recycle to display new content in the banner without deploying?

emclaug2:
> also wondering how we recycle to display new content in the banner without deploying?

I don't think this will be possible without us redeploying our `blui-database`; we won't need to redeploy doc-it for us to update the banner.

emclaug2:
@ektasawant I think the json schema inside the blui-database is good; the issue that @jeffvg pointed out is the most noticeable change.  Other than that, looks good to me.

emclaug2 (approved):

emclaug2 (changes requested):
@ektasawant Sorry I missed this but testing out a failed 'fetch' one more time, 

image

Can you update doc-it so it doesn't display this empty banner if the announcement request fails?

emclaug2:
@ektasawant  Look at the `announcementAppbar.tsx` and `reducers/index.tsx` file in this branch:
https://github.com/brightlayer-ui/doc-it/tree/bug/fix-temp-whitespace

The whitespace (empty app bar) is appearing because we're setting the `showBanner` to be true before it loads in.  This adds a `top: 64` property onto the default banner when it shouldn't be there.  The fix is to default the `showBanner` state to `false`.  We can't show it before the data has been loaded anyways.

ektasawant:
> @ektasawant Look at the `announcementAppbar.tsx` and `reducers/index.tsx` file in this branch: https://github.com/brightlayer-ui/doc-it/tree/bug/fix-temp-whitespace
> 
> The whitespace (empty app bar) is appearing because we're setting the `showBanner` to be true before it loads in. This adds a `top: 64` property onto the default banner when it shouldn't be there. The fix is to default the `showBanner` state to `false`. We can't show it before the data has been loaded anyways.

Thanks @emclaug2 , for fixing this issue. I have updated PR. Please check.

PR #579, opened 2022-09-14
BLUI-1716 Add a component catalog

huayunh:
Fixes BLUI-1716: As Ceviche, I want to know about PX Blue components on doc-it

- Added a "catalog" for a list of components we have
- Made a story on JIRA for migrating this page to dev doc in the future

## Looking for comments on

- Should we list `icon` and `typography` on this page?
- Do pictures make sense to you? Can you identify their corresponding components at a glance?
- For components not available in certain corresponding frameworks, do their tooltip texts make sense?

## Screenshots:
image

## To Test:
- http://localhost:3000/development/components
- Sorry, no firebase hosting on this repository


jeffvg:
I like it looks good. Wondering if we can reuse our card with icons and use the title section for each? maybe too much for this?
image
image

huayunh:
> Wondering if we can reuse our card with icons and use the title section for each? maybe too much for this?

@jeffvg It would not be a significant burden on me; however, I'd like to learn why you would want this change, or what advantage this design brings.

huayunh:
I am also seeking answers to a few particular questions; see the PR description.

ektasawant:
> * Should we list `icon` and `typography` on this page?
> * Do pictures make sense to you? Can you identify their corresponding components at a glance?
> * For components not available in certain corresponding frameworks, do their tooltip texts make sense?

Please check the following observations from my point of view:
Should we list icon and typography on this page? - I guess it is fine since we are showing framework-related components so for RN it can be part of this page.
Do pictures make sense to you? Can you identify their corresponding components at a glance? - Yes
For components not available in certain corresponding frameworks, do their tooltip texts make sense? -  If I didn't read the PR description I couldn't understand I need to hover on strikethrough text. But the tooltip text makes sense.


Not regarding the above. But just wanted to mention 
- heading of this page has a clipboard button but it is invisible.
Screenshot 2022-09-15 at 3 01 52 PM

- heading is not centre aligned.

jeffvg:
> > Wondering if we can reuse our card with icons and use the title section for each? maybe too much for this?
> 
> @jeffvg It would not be a significant burden on me; however, I'd like to learn why you would want this change, or what advantage this design brings.

I think "text" links are ok but visually buttons look better and also the card can hold the requirement of the story bullet point (Include a brief description of what the component is and where to use it)

The strikethrough links to me feel like its not needed when something isn't available. If its not available just don't display the text link...

I don't about the hover statements, not sure what value it has.

image
Are we saying this component is a MUI component and we don't supply one?

huayunh:
thank you both; clipboard icon has been removed, and that hover statement has been fixed

huayunh:
When I designed these, I was thinking that the page would be primarily for new users. They are interested in learning what components we have, and/or if we have a specific component that can come in handy for their project. 
- If we have it, they will click into their framework to learn more about it; 
- if we do not, they may want to know if this is available somewhere else, or if they shall expect it in the future

The second bullet point was the reason I designed that "strike-through" and "tooltip on hover", instead of just hiding a link

jeffvg:
> @jeffvg do you like this better? I didn't add the "short description" part, because I find those images self-explanatory for the "where to use the component" part.
> 
> image

it looks really nice "but"....not playing well at 1024

image

emclaug2 (approved):

PR #587, opened 2022-11-01
Release R25

joebochill:
- All R25 Updates
- Component Catalog
- Icon Guidelines
- Updated release notes & roadmap


huayunh (changes requested):
There should be no changes made to release notes in the history. If you void that rule, I suggest you change all the 50 "PX Blue" references in our release notes too.

https://github.com/brightlayer-ui/doc-it/pull/586

emclaug2:
> There should be no changes made to release notes in the history. If you void that rule, I suggest you change all the 50 "PX Blue" references in our release notes too.
> 
> #586

Looks like a side-effect of some massive copy/paste; we're talking about this right now in office hours.

emclaug2 (approved):

joebochill:
Updating broken links is not fundamentally changing the nature of the release notes...it's just making sure that the links still work. The alternative is to remove the links from old release notes, but I would rather still keep the ability for users to directly link to the relevant item if they are looking through old notes (as long as it's still relevant). I took the links off of things that no longer exist (i.e., DropdownToolbar).

huayunh:
Taking a final walkthrough right now.

> > There should be no changes made to release notes in the history. If you void that rule, I suggest you change all the 50 "PX Blue" references in our release notes too.
> > #586
> 
> Looks like a side-effect of some massive copy/paste; we're talking about this right now in office hours.

That was a deliberate decision to preserve "px blue" when we did the migration.

joebochill:
I've preserved:
- old pxblue npm package links
- react native documentation links

replaced:
- links to specific components to point to the new documentation
- links that point to specific files in github to point to the new org
- links to internal pages in doc it that moved (icon-library)

and removed:
- links to components / orgs that no longer exist
- links to repos that no longer exist

None of the _text_ in any of the release notes has changed.

huayunh (approved):

PR #611, opened 2022-12-19
Tidy up docit mui v5 migration

daileytj:
Fixes BLUI-3671.

- Fix typography table styles
- Fix table of content styles and typography
- Fix some responsiveness issues with table of contents
- Cean up merge conflicts with dev

This should address the remaining issues discussed in the this [PR](https://github.com/brightlayer-ui/doc-it/pull/601)


daileytj:
The link checker is failing both locally and on CI... not sure what we want to do about this since there seem to be a lot of "broken links"

joebochill:
> The link checker is failing both locally and on CI... not sure what we want to do about this since there seem to be a lot of "broken links"

Verify that the 'broken' links actually do work when you open them in a browser. Disable the link checking for CI and we will run it as a manual check before we do any big deployments.

joebochill:
You'll need to fix the imports from all of the content @huayunh added:
image

daileytj:
@joebochill good eye. I had a tough time with that merge.

daileytj:
@joebochill... I must not have run this after I merged this 🤦‍♂️. I'm in there now, I see what is going on. Basically, just have to migrate the stuff that got in early to muiv5. On it!

huayunh:
This spacing right here is missing
image

huayunh:
Table of content should not look like this between 1200px and 1280px

image

huayunh:
This might be my mistake. The design audit page currently has an extra `/` in its URL
image

huayunh:
Dev resources mobile view, these two icons are not aligned
image

huayunh:
Apps built using BLUI, this page's mobile view is not working. Should be a card view

image

huayunh:
Color stripe & list item tag color:
image

huayunh:
Monospaced code inline blocks font size
image

jeffvg:
> 

at 800 width these look ok but when we set to 799 width we drop the github icon
image
image

huayunh:
Color palette, when the screen width is between 900 - 930, the palette is wrapped in a wrong way. Somewhere in the code, we need to change the flexbox style's `justify-content` from `space-between` to `flex-start` 
image

huayunh:
Dark theme, icon library page, the side panel looks... incorrect. The tags field is too bright, and the three developer usage panel is picking up a strange color too.

image

daileytj:
> 1200px and 1280px

This is not being altered at this stage. I spent more than a day going back and forth with multiple people about how this should be done with the new breakpoint defaults in mui5. This is the result of those discussions.

daileytj:
Update: I'll do what I can to address what comes through here, but this issue was scoped to deal with the comments that are currently in the branch this is going to be merged into. If there are issues present here that are currently in the target branch they will need to be addressed in another story.

huayunh:
I think there might be some misunderstanding here. Let me clarify— I am pointing out that there should not be an “active state” on the TOC item. At that viewport width, The TOC is being embedded, instead of fixed. There should not be an active / “current” TOC item being highlighted

daileytj:
> Color stripe & list item tag color. Note that this looks extra wrong in the dark theme, where the color appears extra bright image

Where is this? I'm not seeing this. Is it every example or a select few?

jeffvg:
in all looks like you're covering items needed from migrate to muiv5 PR. 

- also to note setting browser sizes via dev tools versus manually resizing the window...
When using widths 1440, 1024, 768 via the button in dev tools all seems good with exception of "apps built using blui page" and the dev resources page dropping icons....

daileytj:
> in all looks like you're covering items needed from migrate to muiv5 PR.
> 
> * also to note setting browser sizes via dev tools versus manually resizing the window...
>   When using widths 1440, 1024, 768 via the button in dev tools all seems good with exception of "apps built using blui page" and the dev resources page dropping icons....

Just pushed a few fixes. Should cover everything with a thumbs up response above

daileytj:
> Monospaced code inline blocks font size image

Looks like this could have been done on purpose to make all the code blocks 12px. As it stands in dev there are 2 different sizes for code blocks, but a hard coded size for the update. https://brightlayer-ui-dev.github.io/development/environment uses 12px while https://brightlayer-ui-dev.github.io/style/themes#react snippet uses 16px

huayunh:
> Where is this? I'm not seeing this. Is it every example or a select few?

I do not have the computer with me at the moment, but this looks like it comes from the themes page

joebochill:
This is looking great @daileytj — appreciate all of the extra effort on this, including the scope creep.

To weigh in on the code block discussion...the new style defined for dev-docs uses 14 (0.875rem) for inline code. We should be doing the same for doc-it (I would expect that the markdownMapping components should have set this correctly, but there may be something else overriding it like in index.css).

If you've got everything from above handled with the exception of the tags panel in the icon drawer, I would draw a line in the sand there. Anything else (especially if it's (holiday) theme-related) should be a new issue. I'm pretty sure I already logged an issue to fix the themes as they are very very broken from due to some changes in theme structure for MUI 5 and how we do overrides for holiday themes...that is a story of work on its own.

huayunh:
> at 800 width these look ok but when we set to 799 width we drop the github icon
> image
> image

Because there is a “view on GitHub” button below. This was a conscious design choice

huayunh:
I think dev doc used body2 for its paragraph style while docit used body1. We probably want to unify these. I am happy either way … maybe slightly leaning towards body 1

jeffvg:
I pulled down fresh this morning and noticed the warning about pseudo class first-child that would happen once in a few clicks to happening 14 times in one click. If you scroll to the bottom of "apps using blui" page then nav to roadmap page you should see this 14 times

image

jeffvg:
added issue for first-child warnings
https://github.com/brightlayer-ui/doc-it/issues/615

jeffvg:
> 

yeah, was only calling this out because its resizing or padding the "bug" icon and not the build "check icon" throwing off the alignment that you mentioned before here
image

daileytj:
@jeffvg @huayunh Are we good to go on what is here?

jeffvg (approved):

PR #622, opened 2023-01-06
BLUI-3719: material.io -> m2.material.io

huayunh:
- Changed all "Material.io" links to point to "m2.material.io" to ensure that we are always referring to Material Design 2 instead of Material Design 3.


daileytj (approved):
Wise play to lock these down on m2 for now

PR #632, opened 2023-01-18
Temporarily disable markdown link check for contribution guide

daileytj:

- Temporarily disable markdown link check for contribution guide to pass circleci


daileytj:
> What's the purpose of this change? Once the dev -> master merge is finished the link check should pass anyway

Just making sure everything is green. If we want to push it through as is that's fine. I'm also not familiar with all the publishing automation so I'm not sure if there would be any publishing issues related to the build failing in circleci.

huayunh:
If anything is ever going to fail in the dev -> master merge, I would guess it has something to do with firebase & tokens stuff, not with the link checker. I rather not see that link getting bypassed in the link check. I recommend that once we address the icon master branch PR issues, we merge both the icon and docit master PR. If anything goes wrong, we can fix it from there.

PR #673, opened 2023-05-01
Replace codesandbox and stackblitz links with demo site links

daileytj:
Fixes BLUI-3633.

- Update all links to point to our demo projects instead of stackblitz or codesandbox
- Remove all reference to codesandbox or stackblitz outside of recommendations to contributors to provide examples using one of the two tools
-

#### Screenshots / Screen Recording (if applicable)

-

#### To Test:

- Ensure all links work and that there is no reference to codesandbox or stackblitz outside of recommendations to contributors to provide examples using one of the two tools

#### Any specific feedback you are looking for?

-


huayunh (commented):
Overall change looks fine. I noticed this blurb here, and I wonder what you think about it. I can see the benefit of submitting code suggestions via CSB, but if there's a bug, we can't tell if it is a CSB thing.
image

daileytj:
I was on the fence about leaving that in there, but I would prefer to have a running example than not. If we need to reproduce it locally to ensure it isn't specifically codesandbox or stackblitz then so be it. I don't have a strong stance either way, so I opted for not touching it on the first pass. I'm open to suggestions on how we would like to proceed with this. @JeffGreiner-eaton @joebochill any thoughts on the matter?

joebochill:
I would remove the blurb from this page...I don't think it fits contextually even without considering that we're not using them for our examples anymore.

The tools are still useful and I suspect we will still use them...just not as a hosting platform for the code examples. I would consider transitioning this over to the Report Bugs page and having it read something like "When reporting bugs, it can be helpful to submit a link to a Codesandbox or Stackblitz project that demonstrates a reproduction case for the issue...blah blah blah"

huayunh (approved):

